PR #12, opened 2021-08-29
update juce

dromer:
now builds fine on my system at least, check if GH-pages is also pleased ;)


jatinchowdhury18:
Awesome, thanks! I've kicked of the pre-merge builds on this PR, along with a run of the [release pipeline](https://github.com/Chowdhury-DSP/releases/runs/3456907451?check_suite_focus=true) builds.

jatinchowdhury18:
Hmm, so it looks like the release pipeline build are still failing, but since the regular builds passed, I'm going to go ahead and merge anyway.

Thanks!

PR #170, opened 2022-06-16
Corresponding changes to bilinear filters

KaiMikkelsen:
Think I got all the spots where the filter order should be changed, let me know if I missed anything.

Thanks!


KaiMikkelsen:
> Looks good! Just a reminder to push the submodule change first, and then update the submodule here before merging. Thanks!

PR #182, opened 2022-08-08
Cable click add module

KaiMikkelsen:
First Draft of changes allowing a user to right click on a cable, and create a processor within that cables connection. Also includes some refactoring to the Cable and CableView classes.


src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -31,45 +31,46 @@ void CableView::paint (Graphics& g)
...
-            auto startColour = startEditor->getColour();
-            auto endColour = endEditor->getColour();
-            auto levelDB = cable->endProc->getInputLevelDB (cable->endIdx);
-            drawCable (g, startPortLocation.toFloat(), startColour, endPortLocation.toFloat(), endColour, scaleFactor, levelDB);
-        }
-        else if (connectionHelper->cableMouse != nullptr)
+void CableView::mouseDown (const MouseEvent& e)
+{
+    if (e.eventComponent == nullptr)
+        return; // not a valid mouse event
```
jatinchowdhury18, line 57: Interesting... was this something that was happening in your testing/debugging?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -31,45 +31,46 @@ void CableView::paint (Graphics& g)
...
+
+    if (e.mods.isPopupMenu())
+    {
+        for (auto* cable : cables)
         {
-            auto mousePos = connectionHelper->cableMouse->getPosition();
-            const auto nearestInputPort = portLocationHelper->getNearestInputPort (mousePos, cable->startProc);
-            if (nearestInputPort.editor != nullptr && nearestInputPort.isInput && ! portLocationHelper->isInputPortConnected (nearestInputPort))
+            juce::Point clicked (e.getMouseDownX(), e.getMouseDownY());
+            if (cable->contains (clicked)) // Checks hitTest method for cable
```
jatinchowdhury18, line 64: I wonder if it might be better to have this logic live in `Cable::mouseDown()`?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -31,45 +31,46 @@ void CableView::paint (Graphics& g)
...
             {
-                auto endPortLocation = CableViewPortLocationHelper::getPortLocation (nearestInputPort);
-                drawCablePortGlow (g, endPortLocation, scaleFactor);
+                board->popupMenu.showPopupMenu();
```
jatinchowdhury18, line 66: So the issue here is that showing the board popup menu is going to allow the user to replace the cable with any processor, when in this case, the only type of processor that really makes sense is one with a single input and single output.

For example, what happens now if the user tries to replace a cable with a "Tuner" module?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -31,45 +31,46 @@ void CableView::paint (Graphics& g)
...
-            auto levelDB = cable->endProc->getInputLevelDB (cable->endIdx);
-            drawCable (g, startPortLocation.toFloat(), startColour, endPortLocation.toFloat(), endColour, scaleFactor, levelDB);
-        }
-        else if (connectionHelper->cableMouse != nullptr)
+void CableView::mouseDown (const MouseEvent& e)
+{
+    if (e.eventComponent == nullptr)
+        return; // not a valid mouse event
+
+    if (e.mods.isPopupMenu())
```
jatinchowdhury18, line 59: At the moment this won't work on iOS, but that's okay. I can try to implement it sometime next week, or if you're feeling adventurous, you could try to do it using the `chowdsp::PopupMenuHelper`, like in the `BoardComponent`.

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18, line 66: Also, I think more recent JUCE's prefer `showMenuAsync()` for this kind of thing.

jatinchowdhury18 (commented):
Nice work! This is a pretty complicated feature, and it's working really well in my testing. I've left some comments which are mostly just a few minor things, but there's two bigger things that I think should be addressed before merging.

First, it looks like some of my previous commits have been re-added in this PR. I'm not seeing them in the diff, but you still may want to try doing an interactive rebase to make sure they aren't being re-added.

The second thing is more subtle. Basically, I'm worried that the UI and backend logic are becoming a little bit too intertwined, particularly in `ProcessorChainActionHelper`. For reference, it might be useful to look at how the "replace processor" feature is implemented. Basically, when the user clicks "Replace" for a processor, the [`ProcessorStore` is asked to create a menu of eligible processors](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/gui/pedalboard/editors/ProcessorEditor.cpp#L114). In doing so the `ProcessorStore` will make sure that the "action" for each menu item goes through [the proper callback](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/processors/ProcessorStore.cpp#L169), which can then be [implemented in the `ProcessorChainActionHelper`](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/processors/chain/ProcessorChainActionHelper.cpp#L7). That way the action helper doesn't need to store the click state, or the beginning/ending processors, since all that information can be passed through the callback. Does that idea make sense?

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+    auto circle = (Rectangle { cableThickness, cableThickness } * 2.4f * scaleFactor).withCentre (centre);
+    g.setColour (colour);
+    g.fillEllipse (circle);
+
+    g.setColour (Colours::white);
+    g.drawEllipse (circle, portCircleThickness);
+}
+
+void Cable::drawCable (Graphics& g, juce::Point<float> start, Colour startColour, juce::Point<float> end, Colour endColour)
```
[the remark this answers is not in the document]
KaiMikkelsen: Thanks for catching these, I obviously wasn't careful enough with these.

src/gui/pedalboard/cables/Cable.h
```diff
@@ -1,27 +1,59 @@
...
 
+#include "../editors/ProcessorEditor.h"
+#include "CableView.h"
+#include "CubicBezier.h"
 #include "processors/BaseProcessor.h"
+#include <pch.h>
 
-struct Cable
+class BoardComponent;
+class CableView;
```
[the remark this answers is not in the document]
KaiMikkelsen: When I don't have the forward declaration of Cableview, `I get an unknown type name 'CableView' ` error., even with the `#include "CableView.h" `The forward declaration seems like the easy fix here, but is there something else I could try without needing it?

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "CableViewConnectionHelper.h"
+#include "CableViewPortLocationHelper.h"
+
+Cable::Cable (const BoardComponent* comp, const CableView& cv, const ConnectionInfo connection) : startProc (connection.startProc), startIdx (connection.startPort), endProc (connection.endProc), endIdx (connection.endPort), cableView (cv), board (comp), cableColour (0xFFD0592C)
+{
+}
+
+Cable::~Cable() = default;
+
+float Cable::getCableThickness (float levelDB)
```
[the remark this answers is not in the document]
KaiMikkelsen: On this subject, as you can see I opted to have most of the variables needed for cable drawing included as member variables. Do you have any opinion on if having them all as member variables, or allocating them with each call of the paint function would be better?

src/gui/pedalboard/cables/Cable.h
```diff
@@ -1,27 +1,59 @@
...
-    }
+public:
+    Cable (const BoardComponent* comp, const CableView& cv, const ConnectionInfo connection);
+    ~Cable() override;
+
+    auto createCablePath (juce::Point<float> start, juce::Point<float> end, float scaleFactor);
+    float getCableThickness (float levelDB);
+    void drawCableShadow (Graphics& g, float thickness);
+    void drawCableEndCircle (Graphics& g, juce::Point<float> centre, Colour colour);
+    void drawCable (Graphics& g, juce::Point<float> start, Colour startColour, juce::Point<float> end, Colour endColour);
```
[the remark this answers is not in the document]
KaiMikkelsen: Good call, that makes the interface cleaner.

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: This is leftover from the original line checking for various non intended mouse clicks. I haven't used the MouseEvent much so let me know if I misunderstood how that original logic worked.

Screen Shot 2022-08-10 at 3 36 16 PM
.

KaiMikkelsen: Actually reworked this to the original in the latest changes.

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Great idea, really simplified the cableView `mouseDown()` function.

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Mmm this is a good point. The logic I put in `ProcessorChainActionHelper.cpp` should account for any number of inputs and outputs, I tested it with stereo merger, stereo splitter and tuner and they all work as expected (Or at least how I expect). 

The way I was designing this was I figured it was on the user if they wanted to cut off their signal path with a tuner or something similar.

Let me know if I'm misunderstanding your question or we need to change the logic of how it makes connections/displays potential processors to add.

Screen Shot 2022-08-10 at 4 15 14 PM

Screen Shot 2022-08-10 at 4 15 41 PM

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: I modified this with the newest changes. If it still won't work with iOS let me know and I might need some more background on what would need to happen for it to work with iOS. I would be interested to learn what the difference need to be!

src/gui/pedalboard/cables/Cable.h
```diff
@@ -1,27 +1,59 @@
...
+    Colour endColour;
+    juce::Point<float> endPortLocation;
+    float scaleFactor;
+    float levelDB;
+
+    const Colour cableColour; // currently only used for "glow"
+    static constexpr float cableThickness = 5.0f;
+    static constexpr float portCircleThickness = 1.5f;
+
+    static constexpr int getPortDistanceLimit (float scaleFactor) { return int (20.0f * scaleFactor); }
```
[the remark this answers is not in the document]
KaiMikkelsen: Yup, thanks for the catch.

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -78,25 +79,27 @@ void CableView::mouseDown (const MouseEvent& e)
...
     if (nearestPort.isInput)
     {
-        connectionHelper->destroyCable (nearestPort.editor, nearestPort.portIndex);
+        connectionHelper->destroyCable (nearestPort.editor->getProcPtr(), nearestPort.portIndex);
     }
     else
     {
-        connectionHelper->createCable (nearestPort.editor, nearestPort.portIndex, e);
+        connectionHelper->createCable ({ nearestPort.editor->getProcPtr(), nearestPort.portIndex, nullptr, 0 });
+        cableMouse = std::make_unique<MouseEvent> (std::move (e.getEventRelativeTo (this)));
```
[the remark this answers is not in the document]
KaiMikkelsen: Fixed. Thanks for that example, that helped a lot!

src/gui/pedalboard/cables/CableView.h
```diff
@@ -3,39 +3,41 @@
 #include "../editors/ProcessorEditor.h"
 #include "Cable.h"
 
+class Cable;
```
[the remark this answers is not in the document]
KaiMikkelsen: I had same issue as the CableView forward declaration. Getting `'Unknown type name 'Cable''. Could I be missing a Cmake change?

src/gui/pedalboard/cables/CableView.h
```diff
@@ -3,39 +3,41 @@
...
     void paint (Graphics& g) override;
+    void resized() override;
     void mouseDown (const MouseEvent& e) override;
     void mouseDrag (const MouseEvent& e) override;
     void mouseUp (const MouseEvent& e) override;
 
     auto* getConnectionHelper() { return connectionHelper.get(); }
     auto* getPortLocationHelper() { return portLocationHelper.get(); }
     void processorBeingAdded (BaseProcessor* newProc);
+    void processorBeingAddedFromCableClick (BaseProcessor* newProc, Cable* c);
```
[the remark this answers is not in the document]
KaiMikkelsen: Yup, thanks for catching.

src/gui/pedalboard/cables/CableView.h
```diff
@@ -3,39 +3,41 @@
...
-    void setScaleFactor (float newScaleFactor);
-
     struct EditorPort
     {
         ProcessorEditor* editor = nullptr;
         int portIndex = 0;
         bool isInput = false;
     };
 
+    std::unique_ptr<MouseEvent> cableMouse = nullptr;
```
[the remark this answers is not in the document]
KaiMikkelsen: Great point, I added some getters for `isDraggingCable` and `cableMouse->getPosition()`. Let me know if you think there's more that could be done.

src/processors/chain/ProcessorChainActionHelper.h
```diff
@@ -11,6 +11,8 @@ class ProcessorChainActionHelper : private HighResolutionTimer
     void addProcessor (BaseProcessor::Ptr newProc);
     void removeProcessor (BaseProcessor* procToRemove);
     void replaceProcessor (BaseProcessor::Ptr newProc, BaseProcessor* procToReplace);
+    void AddProcessorFromCableClick (BaseProcessor::Ptr newProc);
```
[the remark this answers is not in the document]
KaiMikkelsen: Thanks for catching this

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -31,45 +31,46 @@ void CableView::paint (Graphics& g)
...
+    {
+        for (auto* cable : cables)
         {
-            auto mousePos = connectionHelper->cableMouse->getPosition();
-            const auto nearestInputPort = portLocationHelper->getNearestInputPort (mousePos, cable->startProc);
-            if (nearestInputPort.editor != nullptr && nearestInputPort.isInput && ! portLocationHelper->isInputPortConnected (nearestInputPort))
+            juce::Point clicked (e.getMouseDownX(), e.getMouseDownY());
+            if (cable->contains (clicked)) // Checks hitTest method for cable
             {
-                auto endPortLocation = CableViewPortLocationHelper::getPortLocation (nearestInputPort);
```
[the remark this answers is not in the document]
KaiMikkelsen: Oh interesting, thanks for catching that. I think ideally this would be fixed, even testing changes today, I went to delete a cable and was annoyed when I couldn't click the port when my mouse was over a cable.

I experimented with setting the component in `getNearestPort()` to `board` and that seems to fix the issue. I won't push this change yet as I'm not sure what the side effect of always having that component set to `board` could be, but let me know if you think this could work, or if we should look into a better solution.

Screen Shot 2022-08-10 at 5 41 03 PM

src/processors/ProcessorStore.cpp
```diff
@@ -210,5 +214,12 @@ void ProcessorStore::createProcReplaceList (PopupMenu& menu, int& menuID, BasePr
                    && procInfo.numOutputs == procToReplace->getNumOutputs()
                    && name != procToReplace->getName();
         },
-        procToReplace);
+        procToReplace,
```
KaiMikkelsen, line 217: Passing `nullptr` here. This seems like the least intrusive solution, but let me know if this should be done better.

src/gui/pedalboard/cables/Cable.cpp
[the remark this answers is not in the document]
jatinchowdhury18: I don't think there would be a noticeable performance difference... although I would guess having `levelDB` as a local variable might be a little bit faster than the member variable. That said, I think using the member variable is a bit cleaner, so that's probably the right thing to do for now.

src/gui/pedalboard/cables/Cable.h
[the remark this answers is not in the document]
jatinchowdhury18: Oh interesting... probably some weird "circular include" behaviour. In that case, maybe the best thing to do in both cases would be to use the forward declaration in the header file, and move the `#include` to the .cpp file.

src/gui/pedalboard/cables/CableView.cpp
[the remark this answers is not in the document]
jatinchowdhury18: Yeah, I think that should work... with that change do the ports still glow when the mouse is hovering over the part of the port which is on the module editor? I worry that might get skipped if always using the `board` as the reference component.

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18: Gotcha... I can't think why the `eventComponent` would be null, but I guess it's better to be safe than sorry.

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18: Yeah, so basically instead of using `e.mods.isPopupMenu()` in the `mouseDown()` implementation, you would have to add `chowdsp::PopupMenuHelper` [as a member variable](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/gui/pedalboard/BoardComponent.h#L50), and then configure it [like this](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/gui/pedalboard/BoardComponent.cpp#L63). I won't be able to test on iOS until sometime next week, but if you're able to get it working that way on desktop, I'm guessing it will work fine on iOS as well.

The reason for doing it this way is that there's no way to "right-click" on iOS, so we can configure the plugin to treat a "long-press" as a right-click, which requires some extra logic that is implemented in the helper.

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18: Oh interesting... when I was testing the other day, I got a crash when trying to replace the cable with a tuner, and then doing some undo/redo operations. I'm not 100% what the right user experience is here, but if it's working reliably then I don't see a problem with allowing any number of inputs/outputs for now.

src/processors/chain/ProcessorChainActionHelper.h
```diff
@@ -11,6 +11,7 @@ class ProcessorChainActionHelper : private HighResolutionTimer
     void addProcessor (BaseProcessor::Ptr newProc);
     void removeProcessor (BaseProcessor* procToRemove);
     void replaceProcessor (BaseProcessor::Ptr newProc, BaseProcessor* procToReplace);
+    void addProcessorFromCableClick (BaseProcessor::Ptr newProc, BaseProcessor* startProc, BaseProcessor* endProc);
```
jatinchowdhury18, line 14: I think for this method (and the other similar ones) I'd prefer a name like `replaceConnectionWithProcessor()`, just in case we later choose to allow another interface for interacting with this behaviour rather than clicking on the cables.

src/processors/chain/ProcessorChainActionHelper.cpp
```diff
@@ -46,9 +48,44 @@ void ProcessorChainActionHelper::addProcessor (BaseProcessor::Ptr newProc)
...
 
+void ProcessorChainActionHelper::addProcessorFromCableClick (BaseProcessor::Ptr newProc, BaseProcessor* startProc, BaseProcessor* endProc)
+{
     um->beginNewTransaction();
+
+    // Destroy old cable connection
+    auto minIncomingPorts = jmin (endProc->getNumInputs(), startProc->getNumOutputs());
+    for (int portIdx = 0; portIdx < minIncomingPorts; ++portIdx)
+    {
+        um->perform (new AddOrRemoveConnection (chain, { startProc, portIdx, endProc, portIdx }, true));
```
jatinchowdhury18, line 66: Hmm, so I think this logic might be a little bit too agressive... for example, if the user has the Stereo Splitter/Stereo Merger modules connected to each other with two cables, I think this logic would either remove all the connections (if port 0 is connected to port 0, and port 1 -> port 1), or remove none of the connections (port 0 -> port 1, port 1 -> port 0). Or if only one set of inputs/outputs is connected, I think this logic would try to remove a connection that doesn't exist.

I think a potential solution (which might also make the upstream code a bit cleaner) would be to pass a `Connection` object rather than two `BaseProcessor` objects to this method. That way this logic will know exactly which connection should be removed, and will know with certainty that the connection exists.

src/processors/ProcessorStore.cpp
```diff
@@ -210,5 +214,12 @@ void ProcessorStore::createProcReplaceList (PopupMenu& menu, int& menuID, BasePr
...
+        nullptr,
+        nullptr);
+}
+
+void ProcessorStore::createProcFromCableClickList (PopupMenu& menu, int& menuID, BaseProcessor* cableClickStartProc, BaseProcessor* cableClickEndProc) const
+{
+    createProcListUnfiltered (*this, menu, menuID, nullptr, cableClickStartProc, cableClickEndProc);
```
jatinchowdhury18, line 224: If we did want to enforce that connections should only be allowed to be replaced with 1->1 modules, this would be the place to do it.

src/gui/pedalboard/cables/Cable.h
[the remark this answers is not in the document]
KaiMikkelsen: I tried forward declarations in the headers, and `#include` in the .cpp files, that still gave me some undefined errors unfortunately. 

I changed it so `CableView.h` has no forward declaration but includes `Cable.h`. `Cable.h` keeps the forward declaration but doesn't have an `#include CableView.h`

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Good idea, I implemented a `popupMenu` with a callback in the `Cable` class. I thought about trying to implement the `popupMenu` in `Cableview`, but that would have required some more restructuring, to pass the clicked cable...

Let me know if you think this should work for IOS or if there could be a better way to organize it.

src/processors/ProcessorStore.cpp
KaiMikkelsen: Mmm ok, I looked into default arguments for `createProcListFiltered`, but I get a `Friend declaration specifying a default argument must be a definition` error. For now maybe passing `nullptr` is ok?

src/processors/chain/ProcessorChainActionHelper.h
KaiMikkelsen: Good idea

src/processors/chain/ProcessorChainActionHelper.cpp
KaiMikkelsen: This makes sense, I did some testing with the Splitter and Merger and it did mess with the current logic so thanks for seeing that.

I'm just looking into replacing the two processors with connection, and I think that should make the logic more bulletproof. The only thing I'm not sure about is all the default arguments and `nullptr` checks that would need to be changed. if a `ConnectionInfo` object was passed through instead of the two `BaseProcessor* `'s. For example `createProcListFiltered()` is currently set up to take a startProc and endProc, and we could just pass a `nullptr` when not using those arguments, but I'm not totally sure the best way to approach that problem with a `ConnectionInfo` instead of the two processors.

I'm sure there's a good way to do it, just thought I would ask if you had any recommendations before I looked into it too deeply.

Thanks!

src/processors/ProcessorStore.cpp
jatinchowdhury18: Oh yeah, I forgot that was a friend function. If you want, you could forward declare the function with he default argument (like [this](https://stackoverflow.com/questions/23333949/the-correct-way-to-define-default-argument-for-a-friend-function-in-c)), otherwise, passing nullptr is fine.

src/gui/pedalboard/BoardComponent.cpp
```diff
@@ -165,7 +168,7 @@ void BoardComponent::duplicateProcessor (const ProcessorEditor& editor)
     procChain.getProcStore().duplicateProcessor (*editor.getProcPtr());
 }
 
-void BoardComponent::showNewProcMenu (PopupMenu& menu, PopupMenu::Options& options)
+void BoardComponent::showNewProcMenu (PopupMenu& menu, PopupMenu::Options& options, bool generatedfromCableClick, Cable* clickedCable)
```
jatinchowdhury18, line 171: I think `generatedFromCableClick` might be a redundant argument here, since you could just check`clickedCable == nullptr` and get the same thing?

src/processors/chain/ProcessorChainActionHelper.cpp
jatinchowdhury18: Yeah, that's a good question... I think it should probably start from `createProcFromCableClickList`. Like the signature for that method could be something like `ProcessorStore::createProcFromCableClickList (PopupMenu& menu, int& menuID, ConnectionInfo& connectionInfo)`. Then the connection info could be passed down the callstack. And then you'd probably have to convert between using a reference and pointer whenever there's a default argument that needs to be treated that way.

src/gui/pedalboard/cables/Cable.h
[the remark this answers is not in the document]
jatinchowdhury18: Yeah, that seems like the right fix!

src/gui/pedalboard/BoardComponent.cpp
KaiMikkelsen: Yeah totally, thanks for catching that. I had the same realization yesterday.

src/processors/chain/ProcessorChainActionHelper.cpp
KaiMikkelsen: That helps a lot! I'll look into changing those functions around thanks.

src/processors/ProcessorStore.cpp
KaiMikkelsen: good idea, thanks for that link. I think this could make the code more modifiable going forward.

src/processors/ProcessorStore.cpp
KaiMikkelsen: Good idea, with the new logic this is necessary.

src/gui/pedalboard/cables/CableView.cpp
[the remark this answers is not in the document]
KaiMikkelsen: Yup, they are still glowing using the board. I implemented this change in the newest set of changes.

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -35,6 +35,18 @@ bool Cable::hitTest (int x, int y)
     return false;
 }
 
+ConnectionInfo* Cable::getConnectionInfo()
```
KaiMikkelsen, line 38: I did some thinking about passing a ConnectionInfo object into the ProcessorStore::createProcFromCableClickList () function, like you suggested [here](https://github.com/Chowdhury-DSP/BYOD/pull/182#discussion_r945149583). I thought it might be easier to read if a pointer was just passed right from the start? I could be totally off on this so let me know if there are better reasons to pass a `Cable` then a `ConnectionInfo` object later in the call stack, and I can redo some of these changes. 

I'm not in love with declaring the `ConnectionInfo` here and then assigning the `connectionInfoPtr` member variable to it. For a situation like this I feel like it would make sense to have a `ConnectionInfo` member variable that gets updated as the cable is created and becomes connected. I guess if a `ConnectionInfo` member variable was implemented it could replace the current `startProc`, `endProc`..  member variables.

I thought I would ask what you thought before I went and modified any more but let me know if you have any thoughts on how this should look.

src/processors/ProcessorStore.cpp
```diff
@@ -166,12 +166,14 @@ void createProcListFiltered (const ProcessorStore& store, PopupMenu& menu, int&
...
             {
-                if (procToReplace == nullptr)
-                    store.addProcessorCallback (procDesc.second (store.undoManager));
-                else
+                if (connectionInfo != nullptr)
+                    store.replaceConnectionWithProcessorCallback (procDesc.second (store.undoManager), *connectionInfo);
+                else if (procToReplace != nullptr)
                     store.replaceProcessorCallback (procDesc.second (store.undoManager), procToReplace);
+                else
+                    store.addProcessorCallback (procDesc.second (store.undoManager));
```
jatinchowdhury18, line 176: Nice!

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18: Yeah, your implementation looks great! I'll see if I can give it a go on iOS later this week, but that shouldn't block us from merging this PR.

src/gui/pedalboard/cables/Cable.cpp
jatinchowdhury18: Yeah, passing a pointer throughout is totally fine as well.

There's a couple of interesting points here...
- I'm not sure it makes sense for the `Cable` to own the `ConnectionInfo` as a `std::unique_ptr>`. If `ConnectionInfo` were a big class, it would probably make sense to have this method return a `std::unique_ptr<ConnectionInfo>` and have everything downstream own the pointer. _However_, since `ConnectionInfo` is just a tiny little struct, I think it's fine to just have this method return the struct. Typically for small structs like this the compiler can optimize out any copy operations that would otherwise need to happen when returning a class/struct from a function.
- That said, I think your point about having a `ConnectionInfo` member variable makes even more sense, since at the moment, the `Cable` has a bunch of member variables that basically duplicate the role of `ConnectionInfo` anyway, so probably the way to go is something like:
```cpp
class Cable
{
    // gets updated whenever the cable's connection changes,
    // and replaces current member variables like startProc, endProc
    ConnectionInfo connectionInfo;

    // maybe we don't need this if the member variable is public?
    ConnectionInfo getConnectionInfo() const noexcept
    {
        return connectionInfo;
    }
};
```

jatinchowdhury18: Oh also, even though it probably won't matter, since BYOD is using C++20, we can do [designated initialization](https://www.cppstories.com/2021/designated-init-cpp20/), so we can do this like:
```cpp
ConnectionInfo connection
{
    .startProc = startProc,
    .endProc = endProc,
    .startPort - startIdx,
    .endPort = endIdx,
};
```

src/gui/pedalboard/cables/CableViewConnectionHelper.cpp
```diff
@@ -154,20 +151,23 @@ void CableViewConnectionHelper::releaseCable (const MouseEvent& e)
...
             break;
         }
     }
 
     cableView.repaint();
 }
+
+void CableViewConnectionHelper::clickOnCable (PopupMenu& menu, PopupMenu::Options& options, Cable* clickedCable)
+{
+    board->showNewProcMenu (menu, options, clickedCable->getConnectionInfo());
```
jatinchowdhury18, line 172: board->showNewProcMenu (menu, options, &(clickedCable->connectionInfo));

src/processors/chain/ProcessorChainActionHelper.cpp
```diff
@@ -19,9 +21,25 @@ void ProcessorChainActionHelper::addProcessor (BaseProcessor::Ptr newProc)
...
+        um->beginNewTransaction();
+        um->perform (new AddOrRemoveProcessor (chain, std::move (newProc)));
+    }
+}
 
+void ProcessorChainActionHelper::replaceConnectionWithProcessor (BaseProcessor::Ptr newProc, ConnectionInfo& connectionInfo)
+{
     um->beginNewTransaction();
+
+    um->perform (new AddOrRemoveConnection (chain, { connectionInfo.startProc, 0, connectionInfo.endProc, 0 }, true));
```
jatinchowdhury18, line 35: Use port index if we know it.

jatinchowdhury18:
Looks great! There's one tiny thing in the ActionHelper:
```diff
-    um->perform (new AddOrRemoveConnection (chain, { connectionInfo.startProc, connectionInfo.startPort, newProcRaw, connectionInfo.endPort }));
-    um->perform (new AddOrRemoveConnection (chain, { newProcRaw, connectionInfo.startPort, connectionInfo.endProc, connectionInfo.endPort }));
+    um->perform (new AddOrRemoveConnection (chain, { connectionInfo.startProc, connectionInfo.startPort, newProcRaw, 0 }));
+    um->perform (new AddOrRemoveConnection (chain, { newProcRaw, 0, connectionInfo.endProc, connectionInfo.endPort }));
```

But I'll go ahead and add that, and then merge it. Thanks again!

PR #191, opened 2022-09-14
Initial attempt at recognizing modulation ports

KaiMikkelsen:
First draft at the modulation ports for the Tremolo module. Seems like pretty basic changes, and there are definitely different ways to approach some of the way the ports are labeled, so let me know if you have any ideas for that.

### DIFF
--- a/src/gui/pedalboard/editors/ProcessorEditor.cpp
+++ b/src/gui/pedalboard/editors/ProcessorEditor.cpp
@@ -51,7 +51,7 @@ ProcessorEditor::ProcessorEditor (BaseProcessor& baseProc, ProcessorChain& procs
 
     for (int i = 0; i < baseProc.getNumInputs(); ++i)
     {
-        const auto portType = baseProc.isInputModulationPort(i) ? PortType::modulation : PortType::audio;
+        const auto portType = baseProc.isInputModulationPort (i) ? PortType::modulation : PortType::audio;
         auto newPort = inputPorts.add (std::make_unique<Port> (procUI.backgroundColour, portType));
         newPort->setInputOutput (true);
         addAndMakeVisible (newPort);
@@ -59,7 +59,7 @@ ProcessorEditor::ProcessorEditor (BaseProcessor& baseProc, ProcessorChain& procs
 
     for (int i = 0; i < baseProc.getNumOutputs(); ++i)
     {
-        const auto portType = baseProc.isOutputModulationPort(i) ? PortType::modulation : PortType::audio;
+        const auto portType = baseProc.isOutputModulationPort (i) ? PortType::modulation : PortType::audio;
         auto newPort = outputPorts.add (std::make_unique<Port> (procUI.backgroundColour, portType));
         newPort->setInputOutput (false);
         addAndMakeVisible (newPort);
--- a/src/processors/BaseProcessor.cpp
+++ b/src/processors/BaseProcessor.cpp
@@ -242,7 +242,6 @@ juce::Point<int> BaseProcessor::getPosition (Rectangle<int> parentBounds)
     return (editorPosition * juce::Point { (float) parentBounds.getWidth(), (float) parentBounds.getHeight() }).toInt();
 }
 
-
 void BaseProcessor::routeExternalModulation (const std::initializer_list<int>& inputPorts, const std::initializer_list<int>& outputPorts)
 {
     inputModulationPorts = inputPorts;
@@ -258,4 +257,3 @@ bool BaseProcessor::isOutputModulationPort (int portIndex)
 {
     return outputModulationPorts.contains (portIndex);
 }
-
--- a/src/processors/BaseProcessor.h
+++ b/src/processors/BaseProcessor.h
@@ -106,8 +106,8 @@ public:
 
     const auto& getParameters() const { return AudioProcessor::getParameters(); }
 
-    bool isInputModulationPort(int portIndex);
-    bool isOutputModulationPort(int portIndex);
+    bool isInputModulationPort (int portIndex);
+    bool isOutputModulationPort (int portIndex);
 
 protected:
     virtual void prepare (double sampleRate, int samplesPerBlock) = 0;
--- a/src/processors/other/Tremolo.cpp
+++ b/src/processors/other/Tremolo.cpp
@@ -31,8 +31,7 @@ Tremolo::Tremolo (UndoManager* um) : BaseProcessor ("Tremolo", createParameterLa
     uiOptions.info.description = "A simple tremolo effect.";
     uiOptions.info.authors = StringArray { "Jatin Chowdhury" };
 
-    routeExternalModulation ({ModulationInput}, {ModulationOutput});
-    
+    routeExternalModulation ({ ModulationInput }, { ModulationOutput });
 }
 
 ParamLayout Tremolo::createParameterLayout()
@@ -127,7 +126,7 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
     phaseSmooth.setTargetValue (*rateParam * MathConstants<float>::pi / fs);
     waveSmooth.setTargetValue (*waveParam);
 
-    if (inputsConnected.contains (ModulationInput))// make mono and pass samples through
+    if (inputsConnected.contains (ModulationInput)) // make mono and pass samples through
     {
         // get modulation buffer from input (-1, 1)
         const auto& modInputBuffer = getInputBuffer (ModulationInput);
@@ -140,7 +139,7 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
             modBuffer.applyGain (1.0f / (float) modInputNumChannels);
         }
     }
-    else// create our own modulation signal
+    else // create our own modulation signal
     {
         // fill modulation buffer (-1, 1)
         fillWaveBuffer (modBuffer.getWritePointer (0), numSamples, phase);
--- a/src/processors/other/Tremolo.cpp
+++ b/src/processors/other/Tremolo.cpp
@@ -157,7 +157,7 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
         waveBlock += 0.5f;
 
         // apply depth parameter
-        auto depthVal = std::pow (depthParam->load(), 0.33f);
+        auto depthVal = std::pow (depthParam->getCurrentValue(), 0.33f);
         depthGainSmooth.setTargetValue (depthVal);
         waveBlock.multiplyBy (depthGainSmooth);
         depthAddSmooth.setTargetValue (1.0f - depthVal);
--- a/src/processors/other/Tremolo.h
+++ b/src/processors/other/Tremolo.h
@@ -33,7 +33,7 @@ private:
 
     float fs = 48000.0f;
     float phase = 0.0f;
-    
+
     enum InputPortIndexes
     {
         AudioInput = 0,
--- a/src/processors/utility/StereoMerger.cpp
+++ b/src/processors/utility/StereoMerger.cpp
@@ -37,7 +37,7 @@ void StereoMerger::processAudio (AudioBuffer<float>& buffer)
 
     stereoBuffer.setSize (2, numSamples, false, false, true);
     stereoBuffer.clear();
-    
+
     auto makeBufferMono = [=] (AudioBuffer<float>& b)
     {
         const auto nChannels = b.getNumChannels();


src/gui/pedalboard/editors/Port.cpp
```diff
@@ -1,10 +1,26 @@
...
-Port::Port (const Colour& processorColour) : procColour (processorColour)
+Port::Port (const Colour& processorColour, const PortType type) : procColour (processorColour), portType (type)
 {
 }
 
+Colour Port::getPortColour()
+{
+    if (portType == modulation)
+    {
+        return Colours::rebeccapurple;
```
KaiMikkelsen, line 12: I'm not totally sold on this colour, I was thinking something blue/purplish as those seem to be the go to for modulation.

src/processors/ProcessorStore.cpp
```diff
@@ -53,6 +53,8 @@
 #if BYOD_ENABLE_ADD_ON_MODULES
 #include <AddOnProcessors.h>
 #include <AddOnProcessorStore.h>
+
+
 #endif
```
KaiMikkelsen, line 57: I did a faulty merge in one of my commits and ended up with some of your changes in them. Clang format keeps changing how these are formatted and makes it think there's a change. Is there an easy way to get rid of this change?

src/processors/other/Tremolo.cpp
```diff
@@ -121,26 +123,46 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
-    waveBlock += 0.5f;
-
-    // apply depth parameter
-    auto depthVal = std::pow (depthParam->load(), 0.33f);
-    depthGainSmooth.setTargetValue (depthVal);
-    waveBlock.multiplyBy (depthGainSmooth);
-    depthAddSmooth.setTargetValue (1.0f - depthVal);
-    addSmoothed (waveBlock, depthAddSmooth);
-    filter.process (waveCtx);
+    if (inputsConnected.contains (1))
```
KaiMikkelsen, line 126: I was using the `getInputModulationPort()` here for a bit, but found using 1 and 0 almost made more sense? Let me know if you think there is a cleaner way to do this.

src/processors/other/Tremolo.cpp
```diff
@@ -121,26 +123,46 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
+    {
+        waveBuffer.copyFrom (0, 0, getInputBuffer (1), 0, 0, numSamples);
```
KaiMikkelsen, line 128: Is a copyFrom here the best option, or is there a better way to transfer these samples.

src/processors/other/Tremolo.cpp
jatinchowdhury18: Yeah, I think you're right... the information flow should be like:
- the module informs `getInputModulationPorts()`
- `getInputModulationPorts()` informs the rest of the application

So then the only thing to improve would be getting rid of the magic numbers:
```cpp
class Tremolo
{
    ...
private:
    enum InputPortIndexes
    {
         AudioInput = 0,
         ModulationInput,
    };

    enum OutputPortIndexes
    {
         AudioOutput = 0,
         ModulationOutput,
    };
};

Tremolo::Tremolo()
{
    ...
    routeExternalModulation (ModulationInput, ModulationOutput);
}

void Tremolo::processAudio()
{
    if (inputsConnected.contains (ModulationInput))
        ...
   
    outputBuffers.getReference (ModulationOutput) = ...;
}
```

Actually, we should probably do something similar in the other multi-input/multi-output modules, but that's outside the scope of this PR.

src/processors/BaseProcessor.h
```diff
@@ -161,5 +167,9 @@ class BaseProcessor : private JuceProcWrapper
     StringArray popupMenuParameterIDs;
     OwnedArray<ParameterAttachment> popupMenuParameterAttachments;
 
+    bool modulationPorts = false;
+    int inputModulationPort;
+    int outputModulationPort;
```
jatinchowdhury18, line 172: Since these are uninitialized, other modules seem to show that they sometimes have modulation ports... anyway, this should be fixed with the other changes I had suggested, so probably nothing to worry about.

jatinchowdhury18 (commented):
Looking (and sounding) really cool. There's a few things I made note of, but the overall logic is working great.

One other thing that would be cool is if we could have some of the module's sliders be disabled when the modulation input is connected. Like with the Tremolo module, "Rate" and "Wave" shouldn't do anything when the modulation input is connected, so those sliders should be disabled. Anyway, we can do that in a separate PR if you like.

src/processors/other/Tremolo.cpp
KaiMikkelsen: Good idea, this makes things more clear readability wise

src/processors/other/Tremolo.cpp
KaiMikkelsen: If was collapsed down to mono wouldn't that make processing easier?

src/gui/pedalboard/editors/Port.h
```diff
@@ -2,21 +2,30 @@
...
-    explicit Port (const Colour& processorColour);
+    Port (const Colour& processorColour, const PortType type);
 
     void setInputOutput (bool shouldBeInput) { isInput = shouldBeInput; }
     void setConnectionStatus (bool connectionStatus);
 
     void paint (Graphics& g) override;
 
 private:
+    Colour getPortColour();
```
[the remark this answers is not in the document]
KaiMikkelsen: Good catch thanks

src/processors/other/Tremolo.cpp
jatinchowdhury18: Yeah, I think collapsing to mono makes sense at least for the Tremolo module. I imagine most modulation signals will be mono, so it won't make a difference, but since it's possible to use an audio signal as a modulator, I think summing to mono makes the most sense.

src/gui/pedalboard/editors/ProcessorEditor.cpp
```diff
@@ -51,16 +51,34 @@ ProcessorEditor::ProcessorEditor (BaseProcessor& baseProc, ProcessorChain& procs
...
+            auto newPort = inputPorts.add (std::make_unique<Port> (procUI.backgroundColour, modulation));
+            newPort->setInputOutput (true);
+            addAndMakeVisible (newPort);
+        }
+        else
+        {
+            auto newPort = inputPorts.add (std::make_unique<Port> (procUI.backgroundColour, regular));
+            newPort->setInputOutput (true);
+            addAndMakeVisible (newPort);
+        }
```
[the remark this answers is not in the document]
KaiMikkelsen: Good idea, cool logic

src/processors/BaseProcessor.cpp
```diff
@@ -241,3 +241,29 @@ juce::Point<int> BaseProcessor::getPosition (Rectangle<int> parentBounds)
...
+
+bool BaseProcessor::hasModulationPorts()
+{
+    return modulationPorts;
+}
+
+void BaseProcessor::routeExternalModulation (int inputPort, int outputPort)
+{
+    jassert (numInputs >= 2 && numOutputs >= 2);
+    // If we're making modulation ports make sure there is actually a port for them
```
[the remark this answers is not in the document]
KaiMikkelsen: Would taking this out all together be a good call? I can see cases with no inputs (generator) but any module will always have at least one output or it wouldn't be created, so an assert for a certain value here doesn't really make sense?

src/processors/BaseProcessor.cpp
```diff
@@ -241,3 +241,29 @@ juce::Point<int> BaseProcessor::getPosition (Rectangle<int> parentBounds)
...
+
+    inputModulationPort = inputPort;
+    outputModulationPort = outputPort;
```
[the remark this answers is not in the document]
KaiMikkelsen: Yeah I was thinking of something along these lines earlier as well, and was thinking of some very similar code with a container to hold the ports as well.

I couldn't think of any situations where having multiple modulation ports would be useful, but I would be curious if you have any in mind because there are probably some cool use cases I haven't thought of.

src/processors/other/Tremolo.cpp
```diff
@@ -121,26 +123,46 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
 
-    for (int ch = 0; ch < buffer.getNumChannels(); ++ch)
+    if (! inputsConnected.contains (0)) //If normal input is connected no sounds goes through, only modulation signal
     {
-        auto* x = buffer.getWritePointer (ch);
-        FloatVectorOperations::multiply (x, x, waveBuffer.getReadPointer (0), numSamples);
+        buffer.clear();
     }
+
+    outputBuffers.getReference (0) = &getInputBuffer (0);
```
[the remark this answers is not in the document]
KaiMikkelsen: Yeah, I found that the modulation signal is added to the `buffer` parameter in a situation where I had 2 tremolos.

In the attached picture, using `buffer` instead of `getInputBuffer()`makes the modulation signal be sent to the output instead of just silence/The Input like it should be.

I figured the two signals were being added to the input buffer somewhere in the chain and using the individual `getInputBuffer()`'s seemed like a way of fixing this. Do you think there is a better solution?

Screen Shot 2022-09-22 at 11 14 58 AM

src/processors/other/Tremolo.cpp
```diff
@@ -121,26 +123,46 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
+    else
+    {
+        // fill wave buffer (-1, 1)
+        dsp::AudioBlock<float> waveBlock { waveBuffer };
+        dsp::ProcessContextReplacing<float> waveCtx { waveBlock };
+        fillWaveBuffer (waveBlock.getChannelPointer (0), numSamples, phase);
+
+        // shrink range to (0, 1)
+        waveBlock *= 0.5f;
+        waveBlock += 0.5f;
```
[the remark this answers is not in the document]
KaiMikkelsen: Mmm good point I was wondering if depth should be able to be controlled with each module or just at the modulation source. 

The documentation part seems hard, I guess for now we could keep a note in each module

src/processors/BaseProcessor.cpp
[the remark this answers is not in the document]
jatinchowdhury18: Yeah, it's probably safe to remove it.

src/processors/BaseProcessor.cpp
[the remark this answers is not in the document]
jatinchowdhury18: I was thinking specifically about the chorus module, since it uses a 3-phase LFO... like we could have each phase go to a separate output if we want. That might be a bit extravagant for the chorus, but it would definitely be cool to have the option available if we want to do something like that for future modules.

src/processors/other/Tremolo.cpp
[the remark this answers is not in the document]
jatinchowdhury18: Oh yeah... I should definitely document this somewhere (probably in BaseProcessor.h?), since I forgot about it since the last time I worked with it.

Basically for modules with multiple inputs or outputs, we should ignore the buffer that's passed in through the function argument, and just use `getInputBuffer()` or `getOutputBuffer()` for everything. Sorry about that, I'll make sure to get it documented.

src/processors/other/Tremolo.cpp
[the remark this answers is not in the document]
jatinchowdhury18: Yeah, it's probably best to document that one in BaseProcessor as well.

src/processors/other/Tremolo.cpp
KaiMikkelsen: In the newest changes I stole the `makeBufferMono` lambda from stereoMerger and made it a protected member variable in `BaseProcessor`. does this make sense? Or it could stay as a lambda in both StereoMerger and Tremolo.

src/processors/BaseProcessor.cpp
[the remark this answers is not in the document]
KaiMikkelsen: Ah smart, that would be really cool! Yeah it's not that much work to make it multi modulation input friendly.

src/processors/other/Tremolo.cpp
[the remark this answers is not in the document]
KaiMikkelsen: All good! I was definitely confused by that for a second but luckily the other modules do a good job of using the other buffers.

src/processors/BaseProcessor.cpp
```diff
@@ -241,3 +241,36 @@ juce::Point<int> BaseProcessor::getPosition (Rectangle<int> parentBounds)
 {
     return (editorPosition * juce::Point { (float) parentBounds.getWidth(), (float) parentBounds.getHeight() }).toInt();
 }
+
+void BaseProcessor::makeBufferMono(AudioBuffer<float>& buffer)
+{
+    const auto nChannels = buffer.getNumChannels();
+    const auto numSamples = buffer.getNumSamples();
```
KaiMikkelsen, line 248: Is it safe to assume this argument buffer will have the same number of samples as the other buffers that use this function, or should I include `numSamples` as another argument.

src/gui/pedalboard/editors/Port.h
```diff
@@ -2,21 +2,30 @@
 
 #include <pch.h>
 
+enum PortType
+{
+    audio = 0,
+    modulation,
+};
```
jatinchowdhury18, line 9: I wonder if we should make this an `enum class`? Otherwise, every bit of code that includes this header will have a naming conflict with `audio` and `modulation`.

src/processors/BaseProcessor.h
```diff
@@ -161,5 +172,9 @@ class BaseProcessor : private JuceProcWrapper
     StringArray popupMenuParameterIDs;
     OwnedArray<ParameterAttachment> popupMenuParameterAttachments;
 
+    bool modulationPorts = false;
```
jatinchowdhury18, line 175: I'm not sure we need this member variable.

src/processors/other/Tremolo.cpp
```diff
@@ -118,29 +121,51 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
 {
     const auto numSamples = buffer.getNumSamples();
     waveBuffer.setSize (1, numSamples, false, false, true);
+    modBuffer.setSize (1, numSamples, false, false, true);
```
jatinchowdhury18, line 124: So calling `setSize()` here with the arguments set to avoid reallocating is the right thing to do, but then when does the necessary memory get allocated? There should be another `setSize()` call in the `prepare()` method that allocates the max amount of memory this buffer will need (similar to waveBuffer).

src/processors/other/Tremolo.cpp
```diff
@@ -118,29 +121,51 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
 
-    // fill wave buffer (-1, 1)
-    dsp::AudioBlock<float> waveBlock { waveBuffer };
-    dsp::ProcessContextReplacing<float> waveCtx { waveBlock };
-    fillWaveBuffer (waveBlock.getChannelPointer (0), numSamples, phase);
+    if (inputsConnected.contains (ModulationInput))
+    {
+        // get modulation buffer from input (-1, 1)
+        modBuffer.copyFrom (0, 0, getInputBuffer (ModulationInput), 0, 0, numSamples);
+        makeBufferMono(modBuffer);
```
jatinchowdhury18, line 132: So I'm not sure this logic is exactly right... the way it's written currently, still only the first channel of the input buffer is being used. Really what we want is probably something like:
```cpp
const auto& modInputBuffer = getInputBuffer (ModulationInput);
modBuffer.copyFrom (0, 0, modInputBuffer, 0, 0, numSamples);

if (const auto modInputNumChannels = modInputBuffer.getNumChannels(); modInputNumChannels > 1)
{
    for (int ch = 1; ch < modInputNumChannels; ++ch)
        modBuffer.addFrom (0, 0, modInputBuffer, ch, 0, numSamples);

    modBuffer.applyGain (1.0f / (float) modInputNumChannels);
}
```

src/processors/other/Tremolo.cpp
```diff
@@ -118,29 +121,51 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
+        auto depthVal = std::pow (depthParam->load(), 0.33f);
+        depthGainSmooth.setTargetValue (depthVal);
+        waveBlock.multiplyBy (depthGainSmooth);
+        depthAddSmooth.setTargetValue (1.0f - depthVal);
+        addSmoothed (waveBlock, depthAddSmooth);
+        filter.process (waveCtx);
+
+        for (int ch = 0; ch < buffer.getNumChannels(); ++ch)
+        {
+            auto* x = getInputBuffer (AudioInput).getWritePointer (0);
```
jatinchowdhury18, line 159: I'm not sure this is safe to call before checking `inputsConnected.contains (AudioInput)`? Also, we can probably skip a bunch of the other processing when the input is not connected, to save some CPU.

src/processors/BaseProcessor.cpp
```diff
@@ -241,3 +241,36 @@ juce::Point<int> BaseProcessor::getPosition (Rectangle<int> parentBounds)
 {
     return (editorPosition * juce::Point { (float) parentBounds.getWidth(), (float) parentBounds.getHeight() }).toInt();
 }
+
+void BaseProcessor::makeBufferMono(AudioBuffer<float>& buffer)
```
jatinchowdhury18, line 245: So this could be a static method, but unfortunately I'm not so sure its reuse would be that helpful outside of the stereo merger block.

src/processors/BaseProcessor.cpp
jatinchowdhury18: Yeah, it's probably safe to expect that the buffers will have the same number of samples, but you might want to add a `jassert` just to be sure.

jatinchowdhury18 (commented):
Looking good! Just a few cosmetic changes, and a couple of tweaks in the Tremolo block.

src/processors/BaseProcessor.cpp
KaiMikkelsen: Would the `jassert` be within the `makeBufferMono()` function? To have that wouldn't we need to include a `numSamples` argument for the function? Maybe that's what you were implying or I might be totally missing something.

src/processors/BaseProcessor.h
KaiMikkelsen: Right, thanks for catching that

src/gui/pedalboard/editors/Port.h
KaiMikkelsen: Yeah good call, I think that could save a weird error later on

src/processors/other/Tremolo.cpp
KaiMikkelsen: Ah right, thanks for catching that and the explanation.

src/processors/other/Tremolo.cpp
KaiMikkelsen: Ah I see, I was only copying the 1st channel over before and calling `makeBufferMono`() was redundant. Thanks for catching that

src/processors/other/Tremolo.cpp
KaiMikkelsen: Yeah good call, I updated this in the latest changes.

src/processors/BaseProcessor.cpp
KaiMikkelsen: Yeah definitely, I moved it back to stereo merger

PR #201, opened 2022-09-26
UI redrawing optimizations 

KaiMikkelsen:
From Jatin:
- Redrawing cables makes all the module be redrawn as well
- Cable visualizations off -> no repainting
- Cable viz -> can we generate the paths on a separate thread (use shared Time Slice Thread like oscilloscopes)
- If level isn't changing -> no redrawing cables

### DIFF
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -65,7 +65,6 @@ target_compile_definitions(juce_plugin_modules
         JucePlugin_VersionString="${CMAKE_PROJECT_VERSION}"
         JucePlugin_Name="${CMAKE_PROJECT_NAME}"
         JUCE_MODAL_LOOPS_PERMITTED=1
-        JUCE_ENABLE_REPAINT_DEBUGGING=1
     INTERFACE
         $<TARGET_PROPERTY:juce_plugin_modules,COMPILE_DEFINITIONS>
 )
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -65,6 +65,7 @@ target_compile_definitions(juce_plugin_modules
         JucePlugin_VersionString="${CMAKE_PROJECT_VERSION}"
         JucePlugin_Name="${CMAKE_PROJECT_NAME}"
         JUCE_MODAL_LOOPS_PERMITTED=1
+        JUCE_ENABLE_REPAINT_DEBUGGING=1
     INTERFACE
         $<TARGET_PROPERTY:juce_plugin_modules,COMPILE_DEFINITIONS>
 )
--- a/src/gui/pedalboard/cables/Cable.cpp
+++ b/src/gui/pedalboard/cables/Cable.cpp
@@ -102,7 +102,7 @@ void Cable::repaintIfNeeded (bool force)
 
         auto cableBounds = cablePath.getBounds().toNearestInt();
         cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (2.0f * minCableThickness)));
-        cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (4.0f * minCableThickness) ));
+        cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (4.0f * minCableThickness)));
 
         MessageManager::callAsync ([&, cableBounds]
                                    { repaint (cableBounds); });
--- a/src/gui/pedalboard/cables/Cable.cpp
+++ b/src/gui/pedalboard/cables/Cable.cpp
@@ -101,8 +101,8 @@ void Cable::repaintIfNeeded (bool force)
         cablePath = std::move (createdPath);
 
         auto cableBounds = cablePath.getBounds().toNearestInt();
-        cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (2.0f * minCableThickness)));
-        cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (4.0f * minCableThickness)));
+        cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (4.0f * minCableThickness)));
+        cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (8.0f * minCableThickness)));
 
         MessageManager::callAsync ([&, cableBounds]
                                    { repaint (cableBounds); });
--- a/src/gui/pedalboard/cables/Cable.cpp
+++ b/src/gui/pedalboard/cables/Cable.cpp
@@ -104,8 +104,12 @@ void Cable::repaintIfNeeded (bool force)
         cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (4.0f * minCableThickness)));
         cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (8.0f * minCableThickness)));
 
-        MessageManager::callAsync ([&, cableBounds]
-                                   { repaint (cableBounds); });
+        MessageManager::callAsync (
+            [safeComp = Component::SafePointer (this), cableBounds]
+            {
+                if (auto* comp = safeComp.getComponent())
+                    comp->repaint (cableBounds);
+            });
     };
 
     if (force || connectionInfo.endProc == nullptr)
--- a/src/gui/pedalboard/cables/Cable.cpp
+++ b/src/gui/pedalboard/cables/Cable.cpp
@@ -1,6 +1,5 @@
 #include "Cable.h"
 #include "../BoardComponent.h"
-#include "CableDrawingHelpers.h"
 #include "CableViewConnectionHelper.h"
 #include "CableViewPortLocationHelper.h"
 
@@ -24,7 +23,7 @@ bool Cable::hitTest (int x, int y)
     for (int i = 1; i <= numPointsInPath; ++i)
     {
         auto pointOnPath = bezier.getPointOnCubicBezier ((float) i / (float) numPointsInPath);
-        if (clickedP.getDistanceFrom (pointOnPath) < cablethickness)
+        if (clickedP.getDistanceFrom (pointOnPath) < cableThickness)
         {
             return true;
         }
@@ -33,16 +32,55 @@ bool Cable::hitTest (int x, int y)
     return false;
 }
 
-float Cable::getCableThickness()
+void Cable::updateStartPoint (bool repaintIfMoved)
 {
-    levelDB = jlimit (floorDB, 0.0f, levelDB);
-    auto levelMult = std::pow (jmap (levelDB, floorDB, 0.0f, 0.0f, 1.0f), 0.9f);
-    return cableThickness * (1.0f + 0.9f * levelMult);
+    auto* startEditor = board->findEditorForProcessor (connectionInfo.startProc);
+    jassert (startEditor != nullptr);
+
+    scaleFactor = board->getScaleFactor();
+    startColour = startEditor->getColour();
+    cableThickness = getCableThickness();
+
+    const auto newPortLocation = CableViewPortLocationHelper::getPortLocation ({ startEditor, connectionInfo.startPort, false }).toFloat();
+    if (startPoint.load() != newPortLocation)
+    {
+        startPoint.store (newPortLocation);
+        if (repaintIfMoved)
+            repaintIfNeeded (true);
+    }
 }
 
-auto Cable::createCablePath (juce::Point<float> start, juce::Point<float> end)
+void Cable::updateEndPoint (bool repaintIfMoved)
 {
-    const auto pointOff = portOffset + scaleFactor;
+    if (connectionInfo.endProc != nullptr)
+    {
+        auto* endEditor = board->findEditorForProcessor (connectionInfo.endProc);
+        endColour = endEditor->getColour();
+
+        const auto newPortLocation = CableViewPortLocationHelper::getPortLocation ({ endEditor, connectionInfo.endPort, true }).toFloat();
+        if (endPoint.load() != newPortLocation)
+        {
+            endPoint.store (newPortLocation);
+            if (repaintIfMoved)
+                repaintIfNeeded (true);
+        }
+    }
+    else if (cableView.cableBeingDragged())
+    {
+        endColour = startColour;
+        endPoint = cableView.getCableMousePosition();
+    }
+    else
+    {
+        // when the cable has just been created, we don't have the mouse drag
+        // source yet, so let's just set the end point equal to the start point.
+        endPoint.store (startPoint);
+    }
+}
+
+Path Cable::createCablePath (juce::Point<float> start, juce::Point<float> end, float sf)
+{
+    const auto pointOff = portOffset + sf;
     bezier = CubicBezier (start, start.translated (pointOff, 0.0f), end.translated (-pointOff, 0.0f), end);
     numPointsInPath = (int) start.getDistanceFrom (end) + 1;
     Path bezierPath;
@@ -54,17 +92,58 @@ auto Cable::createCablePath (juce::Point<float> start, juce::Point<float> end)
     return std::move (bezierPath);
 }
 
+void Cable::repaintIfNeeded (bool force)
+{
+    const auto regeneratePath = [this]
+    {
+        auto createdPath = createCablePath (startPoint, endPoint, scaleFactor);
+        ScopedLock sl (pathCrit);
+        cablePath = std::move (createdPath);
+
+        auto cableBounds = cablePath.getBounds().toNearestInt();
+        cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (2.0f * minCableThickness)));
+        cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (4.0f * minCableThickness) ));
+
+        MessageManager::callAsync ([&, cableBounds]
+                                   { repaint (cableBounds); });
+    };
+
+    if (force || connectionInfo.endProc == nullptr)
+    {
+        regeneratePath();
+        return;
+    }
+
+    if (connectionInfo.endProc != nullptr)
+    {
+        auto updatedLevelDB = jlimit (floorDB, 0.0f, connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort));
+        auto levelDifference = std::abs (updatedLevelDB - levelDB);
+        if (std::abs (levelDifference) > 2.0f && levelRange.contains (updatedLevelDB) && levelRange.contains (levelDB))
+        {
+            levelDB = updatedLevelDB;
+            regeneratePath();
+        }
+    }
+}
+
+float Cable::getCableThickness() const
+{
+    auto levelMult = std::pow (jmap (levelDB, floorDB, 0.0f, 0.0f, 1.0f), 0.9f);
+    return minCableThickness * (1.0f + 0.9f * levelMult);
+}
+
 void Cable::drawCableShadow (Graphics& g, float thickness)
 {
+    ScopedLock sl (pathCrit);
     auto cableShadow = Path (cablePath);
     cableShadow.applyTransform (AffineTransform::translation (0.0f, thickness * 0.6f));
     g.setColour (Colours::black.withAlpha (0.3f));
-    g.strokePath (cableShadow, PathStrokeType (cableThickness, PathStrokeType::JointStyle::curved));
+    g.strokePath (cableShadow, PathStrokeType (minCableThickness, PathStrokeType::JointStyle::curved));
 }
 
 void Cable::drawCableEndCircle (Graphics& g, juce::Point<float> centre, Colour colour) const
 {
-    auto circle = (Rectangle { cableThickness, cableThickness } * 2.4f * scaleFactor).withCentre (centre);
+    auto circle = (Rectangle { minCableThickness, minCableThickness } * 2.4f * scaleFactor.load()).withCentre (centre);
     g.setColour (colour);
     g.fillEllipse (circle);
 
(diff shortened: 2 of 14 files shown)


modules/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ target_compile_definitions(juce_plugin_modules
         JucePlugin_VersionString="${CMAKE_PROJECT_VERSION}"
         JucePlugin_Name="${CMAKE_PROJECT_NAME}"
         JUCE_MODAL_LOOPS_PERMITTED=1
+        JUCE_ENABLE_REPAINT_DEBUGGING=1
```
KaiMikkelsen, line 67: Will have to change this before merge

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -102,7 +102,13 @@ void Cable::paint (Graphics& g)
...
         endPortLocation = CableViewPortLocationHelper::getPortLocation ({ endEditor, connectionInfo.endPort, true }).toFloat();
         endColour = endEditor->getColour();
-        levelDB = connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort);
+        auto updatedLevelDb = jlimit (floorDB, 0.0f, connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort));
+        auto levelDifference = std::abs (updatedLevelDb) - std::abs (levelDB);
+        if (std::abs (levelDifference) > 1.0f)
+        {
+            levelDB = updatedLevelDb;
+            repaint();
+        }
```
KaiMikkelsen, line 111: Only update repaint cable when level is changed

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -99,9 +124,26 @@ void CableView::mouseUp (const MouseEvent& e)
 
 void CableView::timerCallback()
 {
```
KaiMikkelsen, line 126: I figured instead of calling `repaint()` every time in the `timerCallback()`, it makes more sense to check if anything needs repainting in timerCallback, then repaint() accordingly. Unfortunately because of how `repaint()` just flags bounds to be repainted instead of immediately going to the function, it seems safest to recheck the repaint() scenario a second time once we actually get to the `paint()` function. Like how `MouseOverClickablePort()` is checked in the `timerCallback` and again in the `paint()` function. Checking twice like this feels inefficient to me but also still eliminates a lot of redundant `repaints()`.

It's very possible there is another way to solve this problem so let me know if you have any ideas!

src/gui/pedalboard/cables/CableView.h
```diff
@@ -52,5 +52,11 @@ class CableView : public Component,
 
     EditorPort nearestPort {};
 
+    bool MouseOverClickablePort();
+
+    bool MouseDraggingOverOutputPort();
```
jatinchowdhury18, line 57: If we could do camel-case for these, that would be fantastic!

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18: Yeah, checking all the ports multiple times isn't great, since we have to do things internally like `getNearestPort()` which has to do a good bit of math. Couldn't we use member variables to cache the values returned by `mouseOverClickablePort()` and company from the `timerCallback()` so we still have those values ready to go in the `paint()` callback?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -99,9 +124,26 @@ void CableView::mouseUp (const MouseEvent& e)
...
-    const auto mousePos = Desktop::getMousePosition() - getScreenPosition();
-    nearestPort = portLocationHelper->getNearestPort (mousePos, board);
-    repaint();
+    if (MouseOverClickablePort())
+    {
+        repaint();
```
jatinchowdhury18, line 129: I wonder if we can avoid repainting the whole view here, like do something like `repaint (boundsAroundPort);`?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -15,31 +15,56 @@ CableView::CableView (BoardComponent* comp) : board (comp)
...
+    if (MouseOverClickablePort())
+    {
+        auto startPortLocation = CableViewPortLocationHelper::getPortLocation (nearestPort);
+        drawCablePortGlow (g, startPortLocation, scaleFactor);
+        portGlowOn = true;
+    }
+    else if (MouseDraggingOverOutputPort())
+    {
+        auto mousePos = cableMouse->getPosition();
+        const auto nearestInputPort = portLocationHelper->getNearestInputPort (mousePos, cables.getLast()->connectionInfo.startProc);
```
jatinchowdhury18, line 63: I wish we could save `nearestInputPort` from `mouseDraggingOverOutputPort()`, since it takes a good bit of work to find it again.

src/gui/pedalboard/cables/Cable.cpp
jatinchowdhury18: Yeah... do you think it would work to move this code to some function like `checkNeedsRepaint()` or something, and then that could get called from `CableView::timerCallback()`?

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -102,7 +102,16 @@ void Cable::paint (Graphics& g)
...
-        levelDB = connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort);
+        auto updatedLevelDb = jlimit (floorDB, 0.0f, connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort));
+        auto levelDifference = std::abs (updatedLevelDb) - std::abs (levelDB);
+        if (std::abs (levelDifference) > 2.0f)
+        {
+            levelDB = updatedLevelDb;
+            auto pathBounds = cablePath.getBounds().toNearestInt();
+            pathBounds.setY (pathBounds.getY() - roundToInt (std::ceil (cableThickness)));
+            pathBounds.setHeight (pathBounds.getHeight() + roundToInt (std::ceil (2.0f * cableThickness)));
+            repaint (pathBounds);
```
jatinchowdhury18, line 113: Nice! I think this is making a big difference.

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -102,7 +102,16 @@ void Cable::paint (Graphics& g)
 
         endPortLocation = CableViewPortLocationHelper::getPortLocation ({ endEditor, connectionInfo.endPort, true }).toFloat();
         endColour = endEditor->getColour();
-        levelDB = connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort);
+        auto updatedLevelDb = jlimit (floorDB, 0.0f, connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort));
+        auto levelDifference = std::abs (updatedLevelDb) - std::abs (levelDB);
+        if (std::abs (levelDifference) > 2.0f)
```
jatinchowdhury18, line 107: I tried playing around with some different thresholds here... the cable visualizations aren't super sensitive, so I think 2 dB is okay here, but we should probably test it out a bit more, to make sure it looks smooth.

We should also probably skip repainting if both `levelDB` and `updatedLevelDb` are below `floorDB`. I also wonder if `floorDB` is a little too low... like the input/output level meters use a floor of -45 dB. Idk, just more stuff to experiment with... and we can always tweak it later.

src/gui/utils/LevelMeterComponent.cpp
```diff
@@ -43,24 +50,22 @@ void LevelMeterComponent::paint (Graphics& g)
...
 
 void LevelMeterComponent::timerCallback()
 {
+    bool needsRepaint = false;
     for (int ch = 0; ch < 2; ++ch)
+    {
         dbLevels[ch] = Decibels::gainToDecibels (levelDetector[ch].processSample (rmsLevels[ch]));
 
-    repaint();
+        if (std::abs (dbLevels[ch] - dbLevelsPrev[ch]) > 0.5f && dbLevels[ch] > minDB && dbLevelsPrev[ch] > minDB)
```
jatinchowdhury18, line 62: I liked your pattern for this, so I figured I'd try something similar with the level meters. I'm trying a threshold of 0.5 dB, but I'm curious if you think it looks a tiny bit choppy?

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -102,7 +102,16 @@ void Cable::paint (Graphics& g)
 
         endPortLocation = CableViewPortLocationHelper::getPortLocation ({ endEditor, connectionInfo.endPort, true }).toFloat();
         endColour = endEditor->getColour();
-        levelDB = connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort);
+        auto updatedLevelDb = jlimit (floorDB, 0.0f, connectionInfo.endProc->getInputLevelDB (connectionInfo.endPort));
+        auto levelDifference = std::abs (updatedLevelDb) - std::abs (levelDB);
```
jatinchowdhury18, line 106:
```suggestion
        auto levelDifference = std::abs (updatedLevelDb - levelDB);
```

jatinchowdhury18 (commented):
These changes look great! Just a few little things to tweak.

src/gui/pedalboard/cables/Cable.cpp
KaiMikkelsen: Yeah good idea I'll look into that

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Yeah I was wondering about that too. Could there be a delay between when `repaint()` is called and when `paint()` is actually run? So by the time `paint()` is run the member variable value could have changed? That was my worry with the member variable, but that's kinda the limits of my `paint()` `repaint()` knowledge so maybe thats not a big worry. I'll test out the member variable and see how it works!

jatinchowdhury18: Well, usually there's not too much delay, but even if there is a delay, the only difference would be that we're painting with a more up-to-date value of the variable, so I think it should be fine.

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Yeah good idea!

src/gui/pedalboard/cables/Cable.cpp
KaiMikkelsen: Mmm, I'm a bit confused with this. Won't the `updatedLevelDb` and `levelDB` always be above the floorDB, because of the `jlimit (floorDB, 0.0f, levelDB)`?

src/gui/utils/LevelMeterComponent.cpp
KaiMikkelsen: Yeah It might be a bit choppier than it was before? I don't necessarily think thats a bad thing if it's adding efficiency? The input and output meters to me are more of a "see if a signal is coming though or not" so as long as they show that, to me thats all I would look for.

src/gui/pedalboard/cables/Cable.cpp
KaiMikkelsen: Good catch, added this in the latest changes.

src/gui/utils/LevelMeterComponent.cpp
jatinchowdhury18: Yeah, that's true... maybe lowering the threshold just a touch wouldn't hurt too much though. It's also a bit hard to tell atm, since everything looks a little bit choppy when `JUCE_ENABLE_REPAINT_DEBUGGING` is turned on :).

KaiMikkelsen: Yeah good point, `JUCE_ENABLE_REPAINT_DEBUGGING` is not a good way to judge UI's

src/gui/pedalboard/cables/Cable.cpp
KaiMikkelsen: I implemented something similar to this in the latest changes. Let me know if you think it's efficient as it can be.

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -112,3 +111,17 @@ void Cable::paint (Graphics& g)
         drawCable (g, startPortLocation, cableView.getCableMousePosition());
     }
 }
+
+Path Cable::pathGeneratorTask::createCablePath (juce::Point<float> start, juce::Point<float> end, float scaleFactor)
```
KaiMikkelsen, line 115: Might need some help/explanation on generating this path with a separate thread

src/gui/utils/LevelMeterComponent.cpp
jatinchowdhury18: I tried with `JUCE_ENABLE_REPAINT_DEBUGGING` turned off and it looked okay to me, so hopefully that's fine!

src/gui/pedalboard/cables/Cable.cpp
jatinchowdhury18: For sure! So first off we probably don' want to use an `AudioUIBackgroundTask` since those are typically reserved for cases where we need the full buffer for whatever we want to do in the UI (like the oscilloscope). Here, probably just using a regular `juce::TimeSliceClient` should be fine.

Then all we need to do is override `TimeSliceClient::useTimeSlice()` to render the cable path(s), and register the client with a `TimeSliceThread`. It would be nice to have the cable drawers on the same `TimeSliceThread` as the Oscilloscope and the Tuner tasks; for that we could make a SharedResourcePointer ([like this](https://github.com/Chowdhury-DSP/chowdsp_utils/blob/master/modules/plugin/chowdsp_plugin_utils/Threads/chowdsp_AudioUIBackgroundTask.h#L36-L41)). Finally , we'll probably need to do some thread synchronization to make sure we're not introducing any data races or similar other threading issues.

One other thought is that we may want to have a single `TimeSliceClient` that renders the paths for all the Cables, rather than one per cable. Maybe it could live in `CableView` and also take care of calling `Cable::checkNeedsRepaint()` or equivalent. Like I'm kind of thinking of it like this:

Audio Thread:
- Generates level data and stores it in atomic variables

TimeSliceThread:
- Loops through all the cables:
  - Cable checks if its level has changed
  - If the level has changed, render a new cable path
  - Use `MessageManager::callAsync()` to trigger a repaint for the cable's bounds

Message Thread:
- Draw the new cable path (should happen automatically in `paint()` method)

Then there's two potential spots where threading issues could creep in, so we'll probably need to protect these spots with `juce::CriticalSection`s.
1. We need to make sure the `Cable` object isn't deleted (on the Message Thread), while being accessed on the TimeSliceThread.
2. We need to make sure the cable `Path` object isn't being simultaneously accessed on the Message and TimeSlice threads. We really don't want to block the MessageThread for long, so it's probably best to create the cable path using a temporary `Path` object, so then we only need to hold the `CriticalSection` while we `std::move()` the temporary Path into the Path object that will be rendered.

KaiMikkelsen: Nice! This helps a lot thanks for the detailed write up. Using the `AudioUIBackgroundTask` didn't really make sense like you said so this clears things up nicely.

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -113,3 +139,18 @@ void CableView::processorBeingRemoved (const BaseProcessor* proc)
...
+    {
+        MessageManager::callAsync ([&]
+                                   { cableView.cables.getLast()->repaint(); });
+    }
+
+    ScopedLock sl (cableView.cableMutex);
+    for (auto* cable : cableView.cables)
+        cable->checkNeedsRepaint();
+
+    return 0;
```
KaiMikkelsen, line 155: Should this be 0?

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -54,17 +92,58 @@ auto Cable::createCablePath (juce::Point<float> start, juce::Point<float> end)
     return std::move (bezierPath);
 }
 
+void Cable::checkNeedsRepaint (bool force)
```
jatinchowdhury18, line 95: I'm not sure `checkNeedsRepaint()` is quite the right name for this method... maybe something more like `repaintIfNeeded()`?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -15,39 +15,54 @@ CableView::CableView (BoardComponent* comp) : board (comp)
 
 CableView::~CableView() = default;
 
-void CableView::paint (Graphics& g)
+bool CableView::mouseOverClickablePort()
 {
-    using namespace CableDrawingHelpers;
-
+    const auto mousePos = Desktop::getMousePosition() - getScreenPosition();
+    nearestPort = portLocationHelper->getNearestPort (mousePos, board);
```
jatinchowdhury18, line 21: I'm having a strange issue where in the Windows VST3 build specifically, this method never recognizes the clickable ports. I think that bug was present before this PR, so I wouldn't worry about trying to fix it now, but I was curious if you'd seen this behaviour at all?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -92,16 +107,33 @@ void CableView::mouseUp (const MouseEvent& e)
...
 
 void CableView::timerCallback()
 {
-    const auto mousePos = Desktop::getMousePosition() - getScreenPosition();
-    nearestPort = portLocationHelper->getNearestPort (mousePos, board);
-    repaint();
+    using namespace CableDrawingHelpers;
+
+    // repaint port glow
+    if (mouseOverClickablePort() || mouseDraggingOverOutputPort())
```
jatinchowdhury18, line 121: It would be good to try to save the value of `mouseOverClickablePort()` so we don't need to call it again in the next if clause.

src/gui/pedalboard/cables/CableView.h
```diff
@@ -34,6 +34,8 @@ class CableView : public Component,
         bool isInput = false;
     };
 
+    bool portGlow = false;
```
jatinchowdhury18, line 37: I think this an be made private?

src/gui/pedalboard/cables/CableView.h
```diff
@@ -51,6 +53,36 @@ class CableView : public Component,
...
+
+        int useTimeSlice() override;
+
+    private:
+        struct TimeSliceThread : juce::TimeSliceThread
+        {
+            TimeSliceThread() : juce::TimeSliceThread ("Cable Drawing Background Thread") {}
+        };
+
+        juce::SharedResourcePointer<TimeSliceThread> sharedTimeSliceThread;
```
jatinchowdhury18, line 82: I wonder if we should make a more global `SharedTimeSliceThread` and have it be used by the cable view as well as the visualizations in the tuner and oscilloscope modules... anyway, we can figure that out later.

src/gui/pedalboard/cables/CableView.h
```diff
@@ -51,6 +53,36 @@ class CableView : public Component,
...
+    {
+    public:
+        explicit PathGeneratorTask (CableView& cv) : cableView (cv)
+        {
+            sharedTimeSliceThread->addTimeSliceClient (this);
+
+            if (! sharedTimeSliceThread->isThreadRunning())
+                sharedTimeSliceThread->startThread();
+        }
+
```
jatinchowdhury18, line 73: We should also have a destructor here that removes the task from the `TimeSliceThread`.

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -113,3 +145,23 @@ void CableView::processorBeingRemoved (const BaseProcessor* proc)
...
+                if (! cableView.cables.isEmpty())
+                    cableView.cables.getLast()->repaint();
+            });
+    }
+
+    ScopedLock sl (cableView.cableMutex);
+    for (auto* cable : cableView.cables)
+        cable->checkNeedsRepaint();
+
+    return 28; // approx. 35 frames / second
```
jatinchowdhury18, line 166: I've set this so wait a few milliseconds between calls so we leave room for other stuff to run the the time slice, but we could probably bring it down a bit if we want to... just depends on what looks best :).

jatinchowdhury18 (approved):
Really nice work! There's a few tiny things, but feel free to merge when you're ready (just remember to turn of the debug repainting flag).

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Mmm I haven't seen that at all, let me know if there's anything I should test to fix that.

src/gui/pedalboard/cables/CableView.cpp
KaiMikkelsen: Good catch, thanks

src/gui/pedalboard/cables/CableView.h
KaiMikkelsen: Mmm yeah that would make sense good idea.

src/gui/pedalboard/cables/CableView.cpp
jatinchowdhury18: Yeah, I'm guessing it's a Windows+VST3 specific thing... I can play around with it later on.

src/gui/pedalboard/cables/CableView.h
KaiMikkelsen: Ah right, thanks for catching that

src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -101,8 +101,8 @@ void Cable::repaintIfNeeded (bool force)
         cablePath = std::move (createdPath);
 
         auto cableBounds = cablePath.getBounds().toNearestInt();
-        cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (2.0f * minCableThickness)));
-        cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (4.0f * minCableThickness)));
+        cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (4.0f * minCableThickness)));
```
KaiMikkelsen, line 104: @jatinchowdhury18 Is there a better way or metric to use for this bound? This is the smallest bound I found where everything was redrawing properly

jatinchowdhury18: Yeah, that's probably as good as we can do... I'll keep thinking about it though.

PR #210, opened 2022-10-07
Flanger Module

KaiMikkelsen:


src/processors/modulation/CleanDelayType.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
KaiMikkelsen, line 1: I thought since this was being used by chorus and flanger now it could be in its own file. I noticed the delay module uses a similar version of this struct but initializes the delay line with a different value 1 << 20 instead of 1 << 18.

Should I look into making this struct more global and we can use it for the delay as well?

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,298 @@
...
+} // namespace
+
+Flanger::Flanger (UndoManager* um) : BaseProcessor ("Flanger",
+                                                    createParameterLayout(),
+                                                    um,
+                                                    magic_enum::enum_count<InputPort>(),
+                                                    magic_enum::enum_count<OutputPort>())
+{
+    using namespace ParameterHelpers;
+    loadParameterPointer (rateParam, vts, "rate");
```
KaiMikkelsen, line 21: It seems like when there is a 5th parameter it's knob is not shown properly. I wondered if we might be changing around some of the parameters anyway so I haven't looked into resizing the module, but let me know what you think of how we can organize all 5 knobs(params)


Screen Shot 2022-10-11 at 12 51 40 PM

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,298 @@
...
+    loadParameterPointer (delayAmountParam, vts, "delayAmount");
+    loadParameterPointer (delayOffsetParam, vts, "delayOffset");
+    loadParameterPointer (fbParam, vts, "feedback");
+    loadParameterPointer (mixParam, vts, "mix");
+    delayTypeParam = vts.getRawParameterValue (delayTypeTag);
+
+    addPopupMenuParameter (delayTypeTag);
+
+    uiOptions.backgroundColour = uiOptions.backgroundColour = Colour (48, 25, 188);
```
KaiMikkelsen, line 30: Blue colour I was experimenting with, definitely some other colour options to go with.

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,298 @@
...
+template <typename DelayArrType>
+void Flanger::processFlanger (AudioBuffer<float>& buffer, DelayArrType& delay)
+{
+    jassert (buffer.getNumChannels() == 2); // always processing in stereo
+    const auto numSamples = buffer.getNumSamples();
+
+    for (int ch = 0; ch < 2; ++ch)
+    {
+        for (int i = 0; i < delaysPerChannel; ++i)
+            delay[ch][i].setFilterFreq (10000.0f);
```
KaiMikkelsen, line 162: I've seen some flangers with a resonance control, would that add anything cool to this one?

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,298 @@
...
+        {
+            auto xIn = std::tanh (x[n] * 0.75f - feedbackState[ch]);
+
+            auto delayValue = delaySmoothSamples[ch].getNextValue();
+            auto delayOffsetValue = delayOffsetSmoothSamples[ch].getNextValue();
+
+            x[n] = 0.0f;
+            for (int i = 0; i < delaysPerChannel; ++i)
+            {
+                float delayAmt = delayValue + delayOffsetValue * 0.5f * (1.0f + 0.95f * LFOData[ch][i][(size_t) n]);
```
KaiMikkelsen, line 186: With the way this is currently set up there will be no effect unless the `delayOffsetValue` != 0. Is that how its supposed to work?

src/processors/modulation/Flanger.cpp
jatinchowdhury18: Maybe... I'm not sure it would be super useful here since the filter frequency isn't available as a user parameter.

jatinchowdhury18 (commented):
This is sounding pretty nice! Just a few thoughts and ideas.

src/processors/modulation/Flanger.cpp
KaiMikkelsen: I was thinking about the Boss pedals and how the chorus is blue and the flanger is purple. In our case the chorus is already purple so maybe some other shade of blue? I can do some experimenting with some other blue/purple colours. Is there any colours you had in mind?

src/processors/modulation/Flanger.h
```diff
@@ -0,0 +1,76 @@
...
+    chowdsp::FloatParameter* rateParam = nullptr;
+    chowdsp::FloatParameter* fbParam = nullptr;
+    chowdsp::FloatParameter* delayAmountParam = nullptr;
+    chowdsp::FloatParameter* delayOffsetParam = nullptr;
+    chowdsp::FloatParameter* mixParam = nullptr;
+    std::atomic<float>* delayTypeParam = nullptr;
+
+    dsp::DryWetMixer<float> dryWetMixer;
+
+    static constexpr int delaysPerChannel = 2;
```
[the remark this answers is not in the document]
KaiMikkelsen: Trying this out works well, and I think does whats intended while saving processing that extra LFO. I don't think the mono sound is really impacted much as well.

src/gui/pedalboard/editors/KnobsComponent.cpp
```diff
@@ -312,6 +312,36 @@ void KnobsComponent::resized()
         return;
     }
 
+    if (totalNumComponents == 5 && sliders.size() == 5)
+    {
+        int compIdx = 0;
+
+        Rectangle<int> bounds[5];
+        const int sWidth = proportionOfWidth (0.215f);
```
KaiMikkelsen, line 320: Could these be more exact?

src/processors/modulation/Flanger.h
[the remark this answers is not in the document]
KaiMikkelsen: I added the LFO + pi/2 to the right like you suggested and I think it sounds good. Let me know if I did this correctly, or if there's further improvements that could be made.

There's quite a few places where the code iterates over the `delaysPerChannel` variable, if we keep this as just 1. Should we change the 2D LFO to be just 1 and stop iterating over each `delayPerChannel`? Or is it worth keeping all that in incase we want to more easily modify that variable in the future?

src/gui/pedalboard/editors/KnobsComponent.cpp
jatinchowdhury18: These seem fine to me. At least when I zoom in/out I don't see any noticeable movement that you might get from rounding errors or something like that.

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,297 @@
...
+            for (int n = 0; n < numSamples; ++n)
+                std::tie (lfoIn[n], lfoOut[n]) = filter.process (lfoIn[n]);
+        };
+
+        auto* modOutData = modOutBuffer.getReadPointer (0);
+
+        FloatVectorOperations::copy (LFOData[0][0].data(), modOutData, numSamples);
+        phaseShiftLFO (LFOData[0][0].data(), LFOData[0][1].data());
+        FloatVectorOperations::copy (LFOData[1][0].data(), LFOData[0][1].data(), numSamples);
+        phaseShiftLFO (LFOData[1][0].data(), LFOData[1][1].data());
```
jatinchowdhury18, line 127: It seems like the plugin crashes here when some external modulation is connected. I think one of the indexes to `LFOData` is out of range?

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+ParamLayout Flanger::createParameterLayout()
+{
+    using namespace ParameterHelpers;
+    auto params = createBaseParams();
+
+    createPercentParameter (params, "rate", "Rate", 0.5f);
+    createTimeMsParameter (params, "delayAmount", "Amount", juce::NormalisableRange { 0.0f, 20.0f }, 2.0f);
+    createTimeMsParameter (params, "delayOffset", "Offset", juce::NormalisableRange { 0.0f, 20.0f }, 1.0f);
```
jatinchowdhury18, line 46: I really like the default parameters that you've chosen! I wonder if we might want to skew the range so that the center value is something like 4-5 milliseconds?

src/processors/modulation/Flanger.cpp
jatinchowdhury18: The new 5-slider layout is looking pretty good!

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,297 @@
...
+    loadParameterPointer (fbParam, vts, "feedback");
+    loadParameterPointer (mixParam, vts, "mix");
+    delayTypeParam = vts.getRawParameterValue (delayTypeTag);
+
+    addPopupMenuParameter (delayTypeTag);
+
+    uiOptions.backgroundColour = uiOptions.backgroundColour = Colour (106, 102, 190);
+    uiOptions.powerColour = Colours::yellow.brighter (0.1f);
+    uiOptions.info.description = "A flanger effect. Use the right-click menu to enable lo-fi mode.";
+    uiOptions.info.authors = StringArray { "Jatin Chowdhury and Kai Mikkelsen" };
```
jatinchowdhury18, line 33:
```suggestion
    uiOptions.info.authors = StringArray { "Kai Mikkelsen", "Jatin Chowdhury" };
```

jatinchowdhury18 (commented):
Looking and sounding really good! I had one comment about the parameter ranges, and I was getting a crash when connecting external modulation, but otherwise, I think this is ready to go!

src/processors/modulation/Flanger.cpp
KaiMikkelsen: Ah good catch, thanks for noticing that

src/processors/modulation/Flanger.cpp
KaiMikkelsen: Good idea!

src/processors/modulation/Flanger.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+            auto& filter = hilbertFilter[filterIndex++];
+            for (int n = 0; n < numSamples; ++n)
+                std::tie (lfoIn[n], lfoOut[n]) = filter.process (lfoIn[n]);
+        };
+
+        auto* modOutData = modOutBuffer.getReadPointer (0);
+
+        FloatVectorOperations::copy (LFOData[0][0].data(), modOutData, numSamples);
+        phaseShiftLFO (LFOData[0][0].data(), LFOData[0][1].data());
```
jatinchowdhury18, line 130:
```suggestion
        phaseShiftLFO (LFOData[0][0].data(), LFOData[1][0].data());
```
I think that puts everything in range? Also, we no longer need two Hilbert filters, since now we're only using one.

KaiMikkelsen: Oh right, thanks for catching that!

jatinchowdhury18 (approved):
Sweet! I think this is good to go!

PR #219, opened 2022-10-25
Tremolo Stereo option

KaiMikkelsen:


src/processors/modulation/Tremolo.cpp
```diff
@@ -128,7 +135,6 @@ void Tremolo::fillWaveBuffer (float* waveBuff, const int numSamples, float& p)
 void Tremolo::processAudio (AudioBuffer<float>& buffer)
 {
     const auto numSamples = buffer.getNumSamples();
-    modOutBuffer.setSize (1, numSamples, false, false, true);
```
jatinchowdhury18, line 131: I think we want to keep this line?

src/processors/modulation/Tremolo.cpp
```diff
@@ -170,13 +179,21 @@ void Tremolo::processAudio (AudioBuffer<float>& buffer)
...
-        for (int ch = 0; ch < numChannels; ++ch)
+        for (int ch = 0; ch < numOutChannels; ++ch)
         {
-            const auto* x = audioInBuffer.getReadPointer (ch);
+            if (stereoMode && ch == 1)
+            {
+                // Flip phase on right LFO
+                auto* xOut = audioOutBuffer.getWritePointer (ch);
+                FloatVectorOperations::negate (xOut, xOut, numSamples);
+                FloatVectorOperations::add (xOut, 1.0f, numSamples);
```
jatinchowdhury18, line 193: So this should work, but it might be better to do this in the loop on line 182-183, so we can save a copy operation? Does that make sense?

jatinchowdhury18 (approved):
Looking good! Just a couple of little things.

src/processors/modulation/Tremolo.cpp
KaiMikkelsen: Mmm, I'm not sure I understand why moving the operation would save a copy? Do we not need to copy either way?

src/processors/modulation/Tremolo.cpp
KaiMikkelsen: Good catch, thanks!

src/processors/modulation/Tremolo.cpp
jatinchowdhury18: Right so basically we could change that for loop to look like:
```cpp
for (int ch = 1; ch < numOutChannels; ++ch)
{
    if (stereoMode)
    {
        // invert modulation signal (no need to copy)
        FloatVectorOperations::negate (audioOutBuffer.getWritePointer (ch), audioOutBuffer.getReadPointer (0), numSamples);
        FloatVectorOperations::add (audioOutBuffer.getWritePointer (ch), 1.0f, numSamples);
    }
    else
    {
        // just copy
        audioOutBuffer.copyFrom (ch, 0, audioOutBuffer, 0, 0, numSamples);
    }
}
```

KaiMikkelsen: Ah I see, I guess I'm still confused how we're able to flip the modulation signal in the `audioOutBuffer`, when my understanding was that the modulation data was written into the `0` channel of the `audioOutBuffer` and  so the other channels of `audioOutBuffer` were empty?

KaiMikkelsen: Oh nice! I see, we can use the read write pointers instead of just the write

PR #220, opened 2022-10-27
Vibrato stereo effect

KaiMikkelsen:


src/processors/modulation/scanner_vibrato/ScannerVibrato.cpp
```diff
@@ -131,25 +136,36 @@ void ScannerVibrato::processAudio (AudioBuffer<float>& buffer)
...
             tapMixTable[i].process (modMixData[0], modMixData[i], numSamples);
 
-        const auto& audioInBuffer = getInputBuffer (AudioInput);
-        const auto numChannels = audioInBuffer.getNumChannels();
-        audioOutBuffer.setSize (numChannels, numSamples, false, false, true);
+        auto& audioInBuffer = getInputBuffer (AudioInput);
+        auto numInChannels = audioInBuffer.getNumChannels();
+        const auto numOutChannels = stereoMode ? 2 : numInChannels;
+
+        if (stereoMode && numInChannels == 1)
```
KaiMikkelsen, line 153: I was having trouble with the mixer when the stereo effect was on, because the input buffer was mono and the output buffer was stereo. I thought having some sort of corrective step was useful to put the input buffer in stereo so the mix would work correctly

src/processors/modulation/scanner_vibrato/ScannerVibrato.cpp
```diff
@@ -167,7 +183,15 @@ void ScannerVibrato::processAudio (AudioBuffer<float>& buffer)
 
             auto* outData = audioOutBuffer.getWritePointer (ch);
             for (int i = 0; i < ScannerVibratoWDF::numTaps; ++i)
+            {
+                if (stereoMode && ch == 1)
```
KaiMikkelsen, line 187: if this the right place to modify the modulation signal to flip the phase?

src/processors/modulation/scanner_vibrato/ScannerVibrato.cpp
```diff
@@ -167,7 +183,15 @@ void ScannerVibrato::processAudio (AudioBuffer<float>& buffer)
...
+                {
+                    FloatVectorOperations::negate (modMixData[i], modMixData[i], numSamples);
```
KaiMikkelsen, line 189: This is how flipping the phase was done in the phaser and tremolo modules, but is it a good way to achieve the flip in this module where the setup is a little different.

jatinchowdhury18: Yeah, `modMixData` should always be in the range [0, 1] at this point, so this should be correct!

src/processors/modulation/scanner_vibrato/ScannerVibrato.cpp
jatinchowdhury18: Yeah, that is tricky. This approach seems good, but I usually try to avoid touching the input buffer if we're doing out-of-place processing, just because other modules might need to use that buffer later.

What if we do something like this:
```cpp
auto& audioInBuffer = getInputBuffer (AudioInput);
const auto numInChannels = audioInBuffer.getNumChannels();
const auto numOutChannels = stereoMode ? 2 : numInChannels;
audioOutBuffer.setSize (numOutChannels, numSamples, false, false, true);
mixer.setWetMixProportion (*mixParam);

if (stereoMode && numInChannels == 1)
{
    for (int ch = 0; ch < numOutChannels; ++ch)
         audioOutBuffer.copyFrom (ch, 0, audioInBuffer, ch % numInChannels, 0, numSamples);

    mixer.pushDrySamples (audioOutBuffer);
}
else
{
     mixer.pushDrySamples (audioInBuffer);
}
```

src/processors/modulation/scanner_vibrato/ScannerVibrato.cpp
```diff
@@ -131,25 +136,44 @@ void ScannerVibrato::processAudio (AudioBuffer<float>& buffer)
...
-        for (int i = ScannerVibratoWDF::numTaps - 1; i >= 0; --i)
-            tapMixTable[i].process (modMixData[0], modMixData[i], numSamples);
+        auto** modMixDataCh0 = modsMixBuffer[0].getArrayOfWritePointers();
+        FloatVectorOperations::add (modMixDataCh0[0], modOutBuffer.getReadPointer (0), 1.0f, numSamples);
+        FloatVectorOperations::multiply (modMixDataCh0[0], depthParam.getSmoothedBuffer(), numSamples); // this also multiplies the signal by 0.5
 
-        const auto& audioInBuffer = getInputBuffer (AudioInput);
-        const auto numChannels = audioInBuffer.getNumChannels();
-        audioOutBuffer.setSize (numChannels, numSamples, false, false, true);
+        auto** modMixDataCh1 = modsMixBuffer[1].getArrayOfWritePointers();
```
KaiMikkelsen, line 146: I couldn't really think of a way to streamline this double declaration of the two modMixes

src/processors/modulation/scanner_vibrato/ScannerVibrato.cpp
```diff
@@ -158,16 +182,45 @@ void ScannerVibrato::processAudio (AudioBuffer<float>& buffer)
...
+                        tapMixTable[i].process (modMixDataCh0[0], modMixDataCh0[i], numSamples);
+                        FloatVectorOperations::addWithMultiply (outData, tapsOutData[i], modMixDataCh0[i], numSamples);
+                    }
+                }
+            }
+            else
+            {
+                for (int i = ScannerVibratoWDF::numTaps - 1; i >= 0; --i)
+                    FloatVectorOperations::addWithMultiply (outData, tapsOutData[i], modMixDataCh0[i], numSamples);
+            }
```
KaiMikkelsen, line 223: This is a pretty long logic statement that could be narrowed down further? do you have any ideas on the best way to cut it down?

KaiMikkelsen:
I think this is getting pretty close, something Isn't quite right with the way the right channel modulation signal is being flipped, it leads to it modulation differently then the left one, but I'm not exactly sure what is going wrong with it?

At 100% and 0% `Width` levels the right and left channels modulate in the same way instead of the different behaviour I was seeing at other width levels, and I'm not really sure why that would be.

jatinchowdhury18:
Sorry, I think some of my suggestions were probably making things a bit more confusing than they should have been. I ended up trying to do something like we had talked about where the modulation data between [0, 1] is stored in an intermediate buffer.

PR #224, opened 2022-10-31
Rotary stereo

KaiMikkelsen:


src/processors/modulation/Rotary.cpp
```diff
@@ -154,12 +164,16 @@ void Rotary::processAudio (AudioBuffer<float>& buffer)
...
+        const auto& audioInBuffer = getInputBuffer (AudioInput);
+        const auto numInChannels = audioInBuffer.getNumChannels();
+        const auto numOutChannels = stereoMode ? 2 : numInChannels;
+
+        audioOutBuffer.setSize (numOutChannels, numSamples, false, false, true);
 
-        for (int ch = 0; ch < numChannels; ++ch)
+        for (int ch = 0; ch < numOutChannels; ++ch)
         {
+            audioOutBuffer.copyFrom (ch, 0, audioInBuffer, ch % numInChannels, 0, numSamples);
```
jatinchowdhury18, line 176: So if the input signal is stereo then the stereo parameter will have no effect? I think that makes sense here.

jatinchowdhury18 (approved):
Looking good!

PR #225, opened 2022-10-31
Replace processor memo

KaiMikkelsen:


KaiMikkelsen:
Do you remember how you got the current gif's generated and recorded? I was trying to create some for the new additions but wasn't able to get the same quality as the ones already in the manual

manual/Manual.md
```diff
@@ -50,6 +53,9 @@ To destroy a cable, click on the input port at the end of the
 connection. Again, the "Undo" button can be used to bring back
 any previously destroyed cables.
 
+### Modulation Ports
+Purple modulation ports are available on all modulation processors. When connected, these ports allow for one processors modulation signal to be passed to another. A processor with a connected modulation input will use that input instead of it's own Rate parameter. This allows multiple modulation processors to be synced to the same rate, creating unique effects possibilities.
```
jatinchowdhury18, line 57:
```suggestion
Purple modulation ports are available on all modulation processors. When connected, these ports allow for one processor's modulation signal to be passed to another. A processor with a connected modulation input will use the modulation signal from that input instead of generating its own. This feature allows multiple modulation processors to be synced to the same modulation signal, creating unique effects possibilities.
```

jatinchowdhury18 (commented):
This text looks really good! It would be nice to have a screen-grab GIF of the cable replacement operation.

I'm not sure it would make sense to try doing a screen-grab of the modulation ports feature... Maybe the best thing to do there is for us to try making a few factory presets that use that feature so folks can see how it works by example.

jatinchowdhury18:
Yeah, the dimensions are fine. It seems like the mouse isn't showing in the GIF? Or maybe that's just on my end. Still, that's probably fine, I think it should be fairly obvious what's going on from the accompanying text.

jatinchowdhury18 (approved):

KaiMikkelsen:
Good catch thanks, I added a new gif with the mouse click.

jatinchowdhury18:
Nice one!

PR #227, opened 2022-11-01
knob component refresh on name change

KaiMikkelsen:


src/gui/pedalboard/editors/KnobsComponent.cpp
```diff
@@ -195,11 +195,33 @@ void KnobsComponent::paint (Graphics& g)
...
 
+    std::unordered_set<String> currentBoxes;
     for (auto* b : boxes)
+    {
         makeName (*b, b->getName());
+        currentBoxes.insert (b->getName());
+    }
+
+    if (needsRepaint (sliderNames, currentSliders) || needsRepaint (boxNames, currentBoxes))
+        repaint();
```
KaiMikkelsen, line 224: Is calling `repaint()` while inside the paint function bad practice?

src/processors/modulation/Panner.cpp
```diff
@@ -304,7 +304,13 @@ bool Panner::getCustomComponents (OwnedArray<Component>& customComps)
...
             leftPanSlider.setVisible (dualPanOn);
 
-            setName (vts.getParameter (dualPanOn ? leftPanTag : mainPanTag)->name);
+            auto newName = vts.getParameter (dualPanOn ? leftPanTag : mainPanTag)->name;
+            if(newName != getName())
+            {
+                setName(newName);
+                if(auto* parent = getParentComponent())
+                    getParentComponent()->repaint();
+            }
```
KaiMikkelsen, line 313: I'm a bit confused on what the `timerCallback` does here. From what I can see whenever the parameter is changed `parameterChanged()` is called, so when would the timerCallback be necessary? Is there an edge case I'm missing.

jatinchowdhury18: Hmm, so for PanSlider2, I remember that we needed the `timerCallback()` in order to update the UI depending on if the module was receiving a stereo or mono input. It's possible that the `timerCallback()` in this class is just a leftover from that, since it looks like the call to `updateSliderVisiblity()` is made redundant by the parameter listener. Have you tried turning off the timer and seeing if anything changes?

jatinchowdhury18 (approved):
Looks good to me! If you want to clean up some of the Timer stuff before merging, feel free to do that as well.

src/processors/modulation/Panner.cpp
KaiMikkelsen: Yeah I haven't seen any changes when getting rid of the timer and doing some testing, I'll take it out and merge!

PR #229, opened 2022-11-03
Generate modulation signal from unconnected processor

KaiMikkelsen:


KaiMikkelsen:
Could be way off on this solution, from the testing I've done it seems to be working as we want but there could be an edge case I haven't thought of.

src/processors/chain/ProcessorChain.cpp
```diff
@@ -175,8 +175,20 @@ void ProcessorChain::processAudio (AudioBuffer<float>& buffer)
             inputBuffer.copyFrom (ch, 0, osBlock.getChannelPointer ((size_t) ch), osNumSamples);
     }
 
-    // run processing chain
+    // process standalone modulation ports
     bool outProcessed = false;
+    for (auto* processor : procs)
+    {
+        auto numImputs = processor->getNumInputConnections();
+
```
jatinchowdhury18, line 183:
```suggestion
```

src/processors/chain/ProcessorChain.cpp
```diff
@@ -175,8 +175,20 @@ void ProcessorChain::processAudio (AudioBuffer<float>& buffer)
...
+        auto isModProcessor = processor->getProcessorType() == Modulation;
```
jatinchowdhury18, line 184: Is this redundant? Like I'm thinking that if this is false, then `modOutputConnected` would also be false, so maybe we don't need to explicitly check this as well? Or maybe we should just move this check into `BaseProcessor::isOutputModulationPortConnected()`?

jatinchowdhury18 (approved):
Yeah, this seems to work well. Let's keep testing this after we merge, just to see if there's a way we can figure out how to break it... but I think this is the solution! Nice work :)

src/processors/chain/ProcessorChain.cpp
KaiMikkelsen: Yeah it seems like we don't need it after doing some testing, good call. The only thing I can think of is checking if it's a modulation processor before looping through the connections to save doing that if it's not a modulation processor?

src/processors/BaseProcessor.cpp
```diff
@@ -286,3 +286,19 @@ bool BaseProcessor::isOutputModulationPort (int portIndex)
 {
     return outputModulationPorts.contains (portIndex);
 }
+
+bool BaseProcessor::isOutputModulationPortConnected()
+{
+    if (getProcessorType() != Modulation)
+        return false;
```
jatinchowdhury18, line 293: Yeah i think this is good.

PR #232, opened 2022-11-04
Update cable positions every move

KaiMikkelsen:


jatinchowdhury18 (commented):
This is looking great! I just ran into one issue that I mentioned in the comments.

KaiMikkelsen:
> This is looking great! I just ran into one issue that I mentioned in the comments.

Weird I can't seem to find that comment?

src/gui/pedalboard/cables/CableView.cpp
```diff
@@ -141,12 +141,6 @@ void CableView::timerCallback()
         portGlow = false;
         repaint (getPortGlowBounds (portToPaint, scaleFactor).toNearestInt());
     }
-
-    for (auto* cable : cables)
-    {
-        cable->updateStartPoint();
-        cable->updateEndPoint();
-    }
```
jatinchowdhury18, line 149: So I'm running into an issue where the cable position doesn't update when it's being dragged, which I think is related to this code being removed. That said, we probably don't need to call this code every time the timer runs, so maybe we could do something like this:
```cpp
if (isDraggingCable)
    updateCablePositions();
```

KaiMikkelsen: Ah right, good idea thanks. I added this line in the latest change.

jatinchowdhury18:
Yeah, this looks great!

PR #233, opened 2022-11-07
Extend Cable Redraw Width

KaiMikkelsen:
Screen Shot 2022-11-07 at 11 07 54 AM

Noticed that the cable bounds were missing some of the outer edges of the cable, and when the cable glow/expand happened these outer edges weren't getting drawn like the rest of the cable.

### DIFF
--- a/src/gui/pedalboard/cables/Cable.cpp
+++ b/src/gui/pedalboard/cables/Cable.cpp
@@ -104,8 +104,8 @@ void Cable::repaintIfNeeded (bool force)
         MessageManager::callAsync (
             [safeComp = Component::SafePointer (this), cableBounds]
             {
-            if (auto* comp = safeComp.getComponent())
-                comp->repaint (cableBounds);
+                if (auto* comp = safeComp.getComponent())
+                    comp->repaint (cableBounds);
             });
     };
 


src/gui/pedalboard/cables/Cable.cpp
```diff
@@ -104,6 +104,9 @@ void Cable::repaintIfNeeded (bool force)
         cableBounds.setY (cableBounds.getY() - roundToInt (std::ceil (4.0f * minCableThickness)));
         cableBounds.setHeight (cableBounds.getHeight() + roundToInt (std::ceil (8.0f * minCableThickness)));
 
+        cableBounds.setX (cableBounds.getX() - roundToInt (std::ceil (minCableThickness)));
+        cableBounds.setWidth (cableBounds.getWidth() + roundToInt (std::ceil (2.0f * minCableThickness)));
+
```
jatinchowdhury18, line 109:
```suggestion
const auto cableBounds = cablePath.getBounds().expanded (std::ceil (minCableThickness), std::ceil (4.0f * minCableThickness).toNearestInt();
```

The review thing won't let me go up to line 103, but I think we could simplify the code like this.

KaiMikkelsen: Oh yeah good idea do the whole bounds in one.

PR #244, opened 2022-11-25
Smooth muff

x31eq:
This adds a parameter to change the value on the capacitors used in the feedback loop for the Muff Drive.  It's required to emulate different Big Muff Pi variants and I think it allows some really good sounds.

The functional change is very simple and I may have done everything else wrong.  Feel free to re-write it as you like.

I've also put a comment in the README about building.  I found it on Debian 5 and it took a while to work out how to disable the Clap build, which I didn't need and fixes the problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [UNRELEASED]
+- Added "Smoothing" parameter for "Muff Drive" module.
+- Added new factory presets.
+- Fixed parameter name changes not showing up in some CLAP hosts.
+
 ## [1.1.0] 2022-11-21
 - Added support for the CLAP plugin format (with parameter modulation).
 - Added modulation ports so modules with internal modulation can share modulation signals.
--- a/README.md
+++ b/README.md
@@ -42,6 +42,14 @@ $ cmake -Bbuild
 $ cmake --build build --parallel 4
 ```
 
+Graham's helpful hint:
+
+```bash
+# This worked for me:
+$ cmake -Bbuild -DBYOD_BUILD_CLAP=OFF
+$ cmake --build build --parallel 4
+```
+
 If you'd like to make an optimized "release" build, it
 is suggested to use some slightly different build commands:
 ```bash
--- a/README.md
+++ b/README.md
@@ -38,21 +38,15 @@ $ cd BYOD
 $ git submodule update --init --recursive
 
 # build with CMake
+# add -DBYOD_BUILD_CLAP=OFF if you get a complaint about the CMake version
 $ cmake -Bbuild
-$ cmake --build build --parallel 4
-```
-
-Graham's helpful hint:
-
-```bash
-# This worked for me:
-$ cmake -Bbuild -DBYOD_BUILD_CLAP=OFF
 $ cmake --build build --parallel 4
 ```
 
 If you'd like to make an optimized "release" build, it
 is suggested to use some slightly different build commands:
 ```bash
+# add -DBYOD_BUILD_CLAP=OFF if you get a complaint about the CMake version
 $ cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
 $ cmake --build build --config Release --parallel 4
 ```
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ $ cd BYOD
 $ git submodule update --init --recursive
 
 # build with CMake
-# add -DBYOD_BUILD_CLAP=OFF if you get a complaint about the CMake version
 $ cmake -Bbuild
 $ cmake --build build --parallel 4
 ```
@@ -46,7 +45,6 @@ $ cmake --build build --parallel 4
 If you'd like to make an optimized "release" build, it
 is suggested to use some slightly different build commands:
 ```bash
-# add -DBYOD_BUILD_CLAP=OFF if you get a complaint about the CMake version
 $ cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
 $ cmake --build build --config Release --parallel 4
 ```
--- a/res/presets/Big Muff (Ram's Head 56).chowpreset
+++ b/res/presets/Big Muff (Ram's Head 56).chowpreset
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<Preset name="Big Muff (Ram's Head 56)" plugin="BYOD" vendor="CHOW" category="Pedals"
+        version="1.1.0">
+  <proc_chain state_plugin_version="1.1.0">
+    <Muff_Drive>
+      <Parameters x_pos="0.2234146296977997" y_pos="0.3075657784938812">
+        <PARAM id="harmonics" value="0.8799999952316284"/>
+        <PARAM id="level" value="0.6499999761581421"/>
+        <PARAM id="n_stages" value="1.0"/>
+        <PARAM id="on_off" value="1.0"/>
+        <PARAM id="sustain" value="0.5"/>
+        <PARAM id="high_q"/>
+        <PARAM id="smoothing" value="0.4550000429153442"/>
+      </Parameters>
+      <port_0 connection_0="1" connection_end_0="0"/>
+    </Muff_Drive>
+    <Muff_Tone>
+      <Parameters x_pos="0.5463414788246155" y_pos="0.320723682641983">
+        <PARAM id="mids" value="0.5"/>
+        <PARAM id="on_off" value="1.0"/>
+        <PARAM id="tone" value="0.5"/>
+        <PARAM id="type" value="1.0"/>
+      </Parameters>
+      <port_0 connection_0="-1" connection_end_0="0"/>
+    </Muff_Tone>
+    <Input>
+      <Parameters x_pos="0.005020080134272575" y_pos="0.3221343755722046">
+        <PARAM id="on_off" value="1.0"/>
+      </Parameters>
+      <port_0 connection_0="0" connection_end_0="0"/>
+    </Input>
+    <Output>
+      <Parameters x_pos="0.8594377636909485" y_pos="0.3221343755722046">
+        <PARAM id="on_off" value="1.0"/>
+      </Parameters>
+    </Output>
+  </proc_chain>
+  <extra_info/>
+</Preset>
--- a/res/presets/Big Muff (Russian).chowpreset
+++ b/res/presets/Big Muff (Russian).chowpreset
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<Preset name="Big Muff (Russian)" plugin="BYOD" vendor="CHOW" category="Pedals"
+        version="1.1.0">
+  <proc_chain state_plugin_version="1.1.0">
+    <Muff_Drive>
+      <Parameters x_pos="0.2234146296977997" y_pos="0.3075657784938812">
+        <PARAM id="harmonics" value="0.8799999952316284"/>
+        <PARAM id="level" value="0.6499999761581421"/>
+        <PARAM id="n_stages" value="1.0"/>
+        <PARAM id="on_off" value="1.0"/>
+        <PARAM id="sustain" value="0.5"/>
+        <PARAM id="high_q"/>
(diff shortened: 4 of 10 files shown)


jatinchowdhury18:
This is super cool, thanks for submitting the PR! I'd like to give it a try with my guitar before merging just get a sense of what it sounds like, but the overall implementation looks solid.

That said, one thing that I might want to change would be to use some sort of parameter smoothing (e.g. [`chowdsp::SmoothedBufferValue`](https://github.com/Chowdhury-DSP/BYOD/blob/9ccff461dac9a02e8eb3771aa8bebe09089755ed/src/processors/modulation/phaser/Phaser8.h#L22)), just to make sure there's no artifacts when quickly changing that parameter. Looks like I'll also have to take a look at the CI pipelines since those are failing because of some token issues.

Thanks for mentioning the CMake version thing as well... The README note is a good idea, but I wonder if I should also make some changes to the CMake configuration, like bumping the CMake required version or adding some warning messages.

Anyway, I'd be curious which Big Muff Pi variants have different feedback capacitors? Maybe that's something we could make a couple of presets out of.

Also, feel free to add yourself to the [AUTHORS](https://github.com/Chowdhury-DSP/BYOD/blob/main/AUTHORS) list, if you'd like!

x31eq:
On 26/11/2022, jatinchowdhury18 ***@***.***> wrote:
> This is super cool, thanks for submitting the PR! I'd like to give it a try
> with my guitar before merging just get a sense of what it sounds like, but
> the overall implementation looks solid.
>
> That said, one thing that I might want to change would be to use some sort
> of parameter smoothing (e.g.
> [`chowdsp::SmoothedBufferValue`](https://github.com/Chowdhury-DSP/BYOD/blob/9ccff461dac9a02e8eb3771aa8bebe09089755ed/src/processors/modulation/phaser/Phaser8.h#L22)),
> just to make sure there's no artifacts when quickly changing that parameter.
> Looks like I'll also have to take a look at the CI pipelines since those are
> failing because of some token issues.

Yes, I think that makes sense, although I'm not sure how to do it.

> Thanks for mentioning the CMake version thing as well... The README note is
> a good idea, but I wonder if I should also make some changes to the CMake
> configuration, like bumping the CMake required version or adding some
> warning messages.

Please don't bump the required version, because that'll break the
whole build on a lot of standard distributions.  Auto-disabling Clap
with a warning would be better.

> Anyway, I'd be curious which Big Muff Pi variants have different feedback
> capacitors? Maybe that's something we could make a couple of presets out
> of.

Okay, yes.  I'm mainly going by this page:

http://www.kitrae.net/music/big_muff_guts.html#Circuit

The original circuit, and so most Triangles, used 500pF capacitors.
That corresponds to 58%, so that's for a Triangle, and also Civil War,
preset.

Some units at different times used 560pF (73%) and that does have
quite a different sound.

Some Ram's Heads used 680pF for the clipping stage which is just
outside the range I set.

The lowest value for official Big Muffs is 430pF (40%) for some early Russians.

I thought I saw a knockoff pedal listed with a much lower capacitance.
I can't find it now so maybe I misread it, in which case it was a
happy accident, because it's worth exploring the lower range.  It
sounds more raw, less like a muff and more like a generic distortion.

Another point to note is that, given real engineering tolerances, the
hardware pedals wouldn't have been made with identical capacitors on
both stages.  For a citation:

    "Many of the components ended up measuring very differently from
the printed values, so much that if a replica was made based on the
printed values alone, it would not sound identical."

    http://www.kitrae.net/music/big_muff_history1RH.html

It isn't worth supporting asymmetric values in the model, though,
because you can always chain two single-stage drives together.  And
this might be something to experiment with.

> Also, feel free to add yourself to the
> [AUTHORS](https://github.com/Chowdhury-DSP/BYOD/blob/main/AUTHORS) list, if
> you'd like!

It's a fairly simple change, so I think being in the git history will be enough.


                       Graham

jatinchowdhury18:
Did some play testing just now and I think this change is sounding really cool! It definitely opens up some different aspects of the sound, especially at the extreme settings.

I just went ahead and added some parameter smoothing. I also changed the parameter to be a "bipolar" percent parameter (i.e. [-100, 100] instead of [0, 100]), so that when the parameter is at 0%, the sound has no change relative to how the module sounded before. Does that seem okay to you? I'm not particularly tied to doing it either way.

> Please don't bump the required version, because that'll break the
whole build on a lot of standard distributions.  Auto-disabling Clap
with a warning would be better.

Ah, that's a good point. I was able to get CMake to show a warning on lower CMake version, but I'll see if I can get auto-disabling working as well.

Thanks for sharing the info about the Big Muff variants as well... I'm not exactly sure what would make sense to do with the presets, since the current "Big Muff" preset has the capacitor at 470pf (closest to the Triangle variant, I got the value from a DAFx paper and I don't remember which variant they were looking at off the top of my head), along with the tone circuit component values from the Ram's Head. So maybe we could make a preset for the Russian variant, but I feel like it would be tough to make a Triangle or Ram's Head preset that would be different _enough_ from the current preset.

> Another point to note is that, given real engineering tolerances, the
hardware pedals wouldn't have been made with identical capacitors on
both stages.

This is a really good point as well and is perfectly in line with the philosophy of BYOD :). Having the cap value as a parameter is definitely a cool way to let folks make use of that sort of variation as well.

x31eq:
On 29/11/2022, jatinchowdhury18 ***@***.***> wrote:
> Did some play testing just now and I think this change is sounding really
> cool! It definitely opens up some different aspects of the sound, especially
> at the extreme settings.

That's good!  It sounds like that to me, but I never know that other
people's ears will be like

> I just went ahead and added some parameter smoothing. I also changed the
> parameter to be a "bipolar" percent parameter (i.e. [-100, 100] instead of
> [0, 100]), so that when the parameter is at 0%, the sound has no change
> relative to how the module sounded before. Does that seem okay to you? I'm
> not particularly tied to doing it either way.

Yes, looks okay.

> Thanks for sharing the info about the Big Muff variants as well... I'm not
> exactly sure what would make sense to do with the presets, since the current
> "Big Muff" preset has the capacitor at 470pf (closest to the Triangle
> variant, I got the value from a DAFx paper and I don't remember which
> variant they were looking at off the top of my head), along with the tone
> circuit component values from the Ram's Head. So maybe we could make a
> preset for the Russian variant, but I feel like it would be tough to make a
> Triangle or Ram's Head preset that would be different _enough_ from the
> current preset.

470pF is the most common capacitor, but 500pF is listed for Triangle.
I'd definitely have a 500pF setting to go with the Triangle filter.
And I can do that regardless of what you set.  I think it's different
enough to be significant although it is subtle.

560pF is the other standard value.  I think 560pF sounds good with low
harmonics and the Ram's Head 1973 filter.  So I'd also add one of
them, but I don't know what to call it.  Probably "Ram's Head 56"
because the "47" specifically identifies 470pF.  For the Russians, it
looks like the Civil War version is specifically identified with
560pF, so that's an easy one.  At least that's what I thought until I
read more: see below

Now I'm checking the circuits here instead of the texts, I can see
that the Ace Tone Fuzz Master FM-3 uses 250pF capacitors:

http://www.bigmuffpage.com/Big_Muff_Pi_versions_schematics_part1.html

Page 2 says Hohner Tri Dirty Booster uses 500pF.  Also, "White Can"
Ram's head uses 560pF, so maybe that's the name.  And some others, in
fact, so you can spend all day at this.

I think it's interesting that they show a 500pF schematic but he says
they really used 470pF.  I think the pedal was designed around 500pF
but the industry standardised on 470pF as the nearest value.  (It's
very difficult to find 500pF capacitors these days.)

Now Page 3

http://www.bigmuffpage.com/Big_Muff_Pi_versions_schematics_part3.html

Red Army 1st Edition is 430pF

Page 4

http://www.bigmuffpage.com/Big_Muff_Pi_versions_schematics_part4.html

It's specifically the Colossal Civil War with 560pF

I'm getting back to work now :)


                     Graham

jatinchowdhury18:
Oh nice, thanks for the extra info on the different cap values! I totally understand spending all day on this sort of thing, since that's exactly what I did with the Muff Tone module :).

How's this for a few different presets:
- "Big Muff" -> same as what's currently in the plugin
- "Big Muff (Triangle)" -> 500pf + Triangle tone circuit
- "Big Muff (Ram's Head 56)" -> 560 pf + Ram's Head '73 tone circuit
- "Big Muff (Russian)" -> 430pf + Russian toen circuit

x31eq:
On 29/11/2022, jatinchowdhury18 ***@***.***> wrote:
> Oh nice, thanks for the extra info on the different cap values! I totally
> understand spending all day on this sort of thing, since that's exactly what
> I did with the Muff Tone module :).
>
> How's this for a few different presets:
> - "Big Muff" -> same as what's currently in the plugin
> - "Big Muff (Triangle)" -> 500pf + Triangle tone circuit
> - "Big Muff (Ram's Head 56)" -> 560 pf + Ram's Head '73 tone circuit
> - "Big Muff (Russian)" -> 430pf + Russian toen circuit

Yes, they look (and I expect will sound) good


             Graham

x31eq:
On 29/11/2022, jatinchowdhury18 ***@***.***> wrote:
> Oh nice, thanks for the extra info on the different cap values! I totally
> understand spending all day on this sort of thing, since that's exactly what
> I did with the Muff Tone module :).

And it was worth it!  It's way better than the two attempts at a Big
Muff emulation my Zoom pedal has.  And you obviously spent a large
amount of time on the Centaur as well.  And what I could say on this:

> How's this for a few different presets:
> - "Big Muff" -> same as what's currently in the plugin
> - "Big Muff (Triangle)" -> 500pf + Triangle tone circuit
> - "Big Muff (Ram's Head 56)" -> 560 pf + Ram's Head '73 tone circuit
> - "Big Muff (Russian)" -> 430pf + Russian toen circuit

* The current one will cover the Ram's Head 47 so that make sense
* You'll have to set the harmonics right.  I don't know how to do this
but I think it'll make a difference for the Russian
* I run with a boost because the level my guitar comes in to the
computer doesn't match the level the plugin's expecting
* I happen to keep my effects settings on github if you want a look at them

https://github.com/x31eq/effects


                           Graham

jatinchowdhury18:
Cool I just added the new presets! If CI passes, and you're okay with everything, then I'll probably go ahead and merge this PR.

> * I run with a boost because the level my guitar comes in to the
computer doesn't match the level the plugin's expecting
> * I happen to keep my effects settings on github if you want a look at them

Yeah, the input level thing is always a bit tricky... the way I usually do it is to have the "Input Gain" (on the bottom bar) set to +9 dB (or a different value if I'm using a guitar other than my own), or if I'm in a DAW maybe I'll have some plugin earlier in the signal chain that gets the level to where I want it to be. The idea with the "Input Gain" is that it's not saved/loaded as part of any presets, so I can flip through presets without having to worry about always adding some extra input gain.

Yeah, your effects settings look pretty nice! If you'd like to have any of them included as factory presets, just let me know. (We can do that in a new PR).

x31eq:
On 03/12/2022, jatinchowdhury18 ***@***.***> wrote:
> Cool I just added the new presets! If CI passes, and you're okay with
> everything, then I'll probably go ahead and merge this PR.

I see this got merged, so thanks for that

>> * I run with a boost because the level my guitar comes in to the
> computer doesn't match the level the plugin's expecting
>> * I happen to keep my effects settings on github if you want a look at
>> them
>
> Yeah, the input level thing is always a bit tricky... the way I usually do
> it is to have the "Input Gain" (on the bottom bar) set to +9 dB (or a
> different value if I'm using a guitar other than my own), or if I'm in a DAW
> maybe I'll have some plugin earlier in the signal chain that gets the level
> to where I want it to be. The idea with the "Input Gain" is that it's not
> saved/loaded as part of any presets, so I can flip through presets without
> having to worry about always adding some extra input gain.

Okay, right, that does make sense.  So I've set it to the full 18dB
and revised all my presets accordingly.

> Yeah, your effects settings look pretty nice! If you'd like to have any of
> them included as factory presets, just let me know. (We can do that in a new
> PR).

I don't think there's any need for it, they aren't anything special.
Although having a Centaur followed by a muff does work so you can take
that idea.  I might do some more effects one day, or I might not, I
don't know what'll occupy my attention.  But I'll certainly keep using
BYOD, it's a good set of effects.

I did make some videos already as well https://youtu.be/bx9yIOpC4Tw


                     Graham

PR #251, opened 2022-12-20
Fix crash when loading plugin state in iOS GarageBand

### LINKED ISSUE #247: [BUG] AUV3/Official app version crashes GarageBand when using more than once.
I’ll start this off with this, I don’t know how to code lol but I’m writing this as a users perspective. From what I’ve found out with my tunes, I can’t load some of them up because of this massive and detrimental bug. Basically if you load up a patch and then load up BYOD it would be alright. But once you’ve made a chain and your own plugin then make a new one/duplicate the track, it would crash the whole app booting you back to the Home Screen. Hopefully someone has since found the problem but it’s been a pain to know that the recent update has effectively broke the app.
I’ve done a screen record to show an example: https://www.dropbox.com/s/ddcqv145qr6v2h7/2022-12-06%209.48.59%20pm.mp4?dl=0

jatinchowdhury18:
Should fix #247.

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -20,11 +20,9 @@ jobs:
     steps:
     - name: Install Linux Deps
       run: |
-        wget https://apt.llvm.org/llvm.sh
-        chmod +x llvm.sh
-        sudo ./llvm.sh 12
-        sudo apt-get install clang-format-12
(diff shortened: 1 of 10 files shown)


sks2002Official:
Hi, could you send a TestFlight link for me please so I could test it out to see if that fixed it?

PR #252, opened 2022-12-31
Muff Clipper

x31eq:
Muff Drive copied and extended to have some asymmetric diode modelling.  Made single-stage so that it can be chained with different settings to give more flexibility.  A couple of presets as examples

### DIFF
--- a/res/presets/Gainful Clipper.chowpreset
+++ b/res/presets/Gainful Clipper.chowpreset
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<Preset name="Gainful Clipper" plugin="BYOD" vendor="User" category="Pedals"
-        version="1.1.1" preset_file="/home/graham/music/effects/User/Pedals/Gainful Clipper.chowpreset">
-  <proc_chain state_plugin_version="1.1.1">
+<Preset name="Gainful Clipper" plugin="BYOD" vendor="CHOW" category="Pedals"
+        version="1.1.0">
+  <proc_chain state_plugin_version="1.1.0">
     <Muff_Tone>
       <Parameters x_pos="0.6112957000732422" y_pos="0.2140151560306549">
         <PARAM id="mids" value="0.3120000660419464"/>
--- a/res/presets/Gainful Clipper.chowpreset
+++ b/res/presets/Gainful Clipper.chowpreset
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<Preset name="Gainful Clipper" plugin="BYOD" vendor="CHOW" category="Pedals"
+<Preset name="Gainful Clipper" plugin="BYOD" vendor="x31eq" category="Pedals"
         version="1.1.0">
   <proc_chain state_plugin_version="1.1.0">
     <Muff_Tone>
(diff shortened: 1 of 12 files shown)


res/presets/Gainful Clipper.chowpreset
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<Preset name="Gainful Clipper" plugin="BYOD" vendor="CHOW" category="Pedals"
```
jatinchowdhury18, line 3: If you'd prefer to put your own name as the vendor, that would be cool as well. Totally up to you!

src/processors/drive/muff_clipper/MuffClipper.cpp
```diff
@@ -0,0 +1,192 @@
...
+{
+    using namespace ParameterHelpers;
+
+    auto params = createBaseParams();
+
+    createPercentParameter (params, "sustain", "Gain", 0.5f);
+    createPercentParameter (params, "harmonics", "Harm.", 0.65f);
+    createBipolarPercentParameter (params, "smoothing", "Smooth", 0.0f);
+    createBipolarPercentParameter (params, "clip1", "+Clip", 0.0f);
+    createBipolarPercentParameter (params, "clip2", "-Clip", 0.0f);
```
jatinchowdhury18, line 45: From looking at the output in an oscilloscope I think these names might be flipped?

src/processors/drive/muff_clipper/MuffClipper.cpp
```diff
@@ -0,0 +1,192 @@
...
+    clip2Param.setParameterHandle (getParameterPointer<chowdsp::FloatParameter*> (vts, "clip2"));
+    smoothingParam.setParameterHandle (getParameterPointer<chowdsp::FloatParameter*> (vts, "smoothing"));
+    hiQParam = vts.getRawParameterValue ("high_q");
+
+    addPopupMenuParameter ("high_q");
+
+    uiOptions.backgroundColour = Colours::darkgrey.brighter (0.3f).withRotatedHue (0.2f);
+    uiOptions.powerColour = Colours::red.brighter (0.15f);
+    uiOptions.info.description = "Fuzz effect based on a single drive stage from the Electro-Harmonix Big Muff Pi.";
+    uiOptions.info.authors = StringArray { "Jatin Chowdhury", "Graham Breed" };
```
jatinchowdhury18, line 32: I would have put your name first, but that's totally your call as well.

jatinchowdhury18 (approved):
Great stuff, this is sounding fantastic! The Gainful Clipper preset is an instant favorite :).

I made a pass doing a little bit of cleanup and optimizations, so you may want to glance at that just in case I messed something up.

x31eq:
On 07/01/2023, jatinchowdhury18 ***@***.***> wrote:
> @jatinchowdhury18 approved this pull request.
>
> Great stuff, this is sounding fantastic! The Gainful Clipper preset is an
> instant favorite :).
>
> I made a pass doing a little bit of cleanup and optimizations, so you may
> want to glance at that just in case I messed something up.

Okay.  I could put my name as vendor but I thought the convention
would be that the presets were coming from the factory.  The preset as
a whole is still mostly your code, so your name should be first there.

I think the issue with the +/- will be that a single stage inverts the
signal, and that's corrected for at the end.  So it makes sense that
they were wrong and I've reversed the names.

Your changes look fine to me.  I was getting failures from github that
might have been consistent with main, but at least I didn't understand
them.


                 Graham

jatinchowdhury18:
Sure thing! Yeah, no worries about the GitHub Actions failure, that's just a code-formatting thing that isn't too important for right now.

Anyway, merging now, and I'll update the Changelog a little later. Thanks again for the contribution!

PR #272, opened 2023-04-05
Add virtual analog ladder filter processor

butchwarns:
This PR adds a new processor to the tone category.

The "Ladder Filter" is a dual resonating four-pole filter modeled after the classic transistor-ladder.
The model was created by the topology-preserving transform (TPT) method, as per Zavalishin's "The Art of VA Filter Design". A saturating nonlinearity can be used to add some grit to the output. Both filters are capable of self-oscillation at high resonance settings, and allow for smooth modulation of the cutoff frequencies.

Please let me know if your expertise dictates anything needs changing, or if you see room for improvement!

-Butch


src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
```diff
@@ -0,0 +1,68 @@
...
+            x = hp[channel].process (x);
+            x = lp[channel].process (x);
+
+            // Appply drive
+            x *= gain;
+
+            // Saturate (static nonlinearity) and blend using drive param
+            x = p.drive_normalized() * fast_tanh_2 (x) + (1.0 - p.drive_normalized()) * x;
+        }
+    }
```
jatinchowdhury18, line 67: So I've been noticing a bit of "crackling" when moving the sliders around, and I think that's coming from the fact the the parameter smoothing seems to be happening per-block rather than per-sample.

I think what I'd recommend is to use the `chowdsp::SmoothedBufferValue` for smoothing the parameters (the new [Flapjack](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/processors/drive/flapjack/Flapjack.h#L18-L20) processor has an example of this). That way we can know when the parameter needs to be smoothed, and when it's staying constant. `SmoothedBufferValue` also has a way of setting a "mapping function" which can be used to apply "skews" to the parameters when needed.

From there, we can re-write this top-level processing block in a way that should help both the smoothing, and the CPU performance:
```cpp
const auto num_samples = buffer.getNumSamples();

// Set up parameter smoothers
hp_cutoff_smooth.process (numSamples); // hp_cutoff_smooth is a chowdsp::SmoothedBufferValue
hp_resonance_smooth.process (numSamples);
// do the rest of the parameters...

for (int channel = 0; channel < buffer.getNumChannels(); ++channel)
{
    auto* data = buffer.getWritePointer (channel);

    // process hpf
    if (hp_cutoff_smooth.isSmoothing() || hp_resonance_smooth.isSmoothing())
    {
        // if we need parameter smoothing...
        const auto* cutoff_smooth = hp_cutoff_smooth.getSmoothedBuffer();
        const auto* resonance_smooth = hp_resonance_smooth.getSmoothedBuffer();
        for (int n = 0; n < num_samples; ++n)
        {
            hp[channel].set_cutoff (cutoff_smooth[n]);
            hp[channel].set_resonance (resonance_smooth[n]);
            data[n] = hp[channel].process (data[n]);
        }
    }
    else
    {
        // no need for parameter smoothing
       hp[channel].set_cutoff (hp_cutoff_smooth.getCurrentValue());
       hp[channel].set_resonance (hp_resonance_smooth.getCurrentValue());
        for (int n = 0; n < num_samples; ++n)
        {
            data[n] = hp[channel].process (data[n]);
        }
    }

    // here we would do the same thing for the lowpass filter

    // apply drive
    juce::FloatVectorOperations::multiply (data, drive.getSmoothedBuffer(), num_samples);

    // saturate and blend
    const auto* drive_norm_smooth = drive_normalize_smooth.getSmoothedBuffer();
    for (int n = 0; n < num_samples; ++n)
        data[n] = drive_norm_smooth[n] * fast_tanh (data[n]) + (1.0f - drive_norm_smooth[n]) * data[n];
}
```

Anyway, that's a rough outline, but let me know if you have any questions or other ideas for how to do this sort of thing.

src/processors/tone/ladder_filter/HighPassOnePole.h
```diff
@@ -0,0 +1,42 @@
...
+    // (g2 is calculated when setting the cutoff of the full ladder structure)
+    void set_g2 (const double _g2);
+
+    // Get current contents of the internal state register
+    double get_state() const;
+
+    //==============================================================================
+
+    // Process a single audio sample
+    double process (double x);
```
jatinchowdhury18, line 34: I wonder if we might benefit from some of these per-sample processing functions being marked `inline`... Compilers these days are probably smart enough to inline them anyway, but it might be worth a try.

jatinchowdhury18 (commented):
Great work! Overall it's sounding good, and seems quite stable.

I had a couple of notes on some code organization, and some things related to parameter smoothing, definitely let me know if you have any questions.

One other thing I was thinking is that the self-oscillation could be a little dangerous for some users who would start playing around with the filter, not expecting it to be able to do that... I wonder if we might want to have two different "ranges" for the "Resonance" parameters, where the "normal" range doesn't self-oscillate, and the "extended" range does. Then we could have some other parameter (accessible from the settings menu) that could turn on the extended range. On the other hand maybe that's being a little too cautious, since the self-oscillation can be quite fun. Anyway, I'm curious to hear your thoughts.

src/processors/tone/ladder_filter/HighPassOnePole.h
butchwarns: Consider this done.

src/processors/tone/ladder_filter/utility.h
```diff
@@ -0,0 +1,99 @@
+#pragma once
+
+#include <cmath>
+
+//==============================================================================
```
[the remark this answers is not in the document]
butchwarns: Consider this done.

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
butchwarns: Good point! I also think the smoothing can be improved. I will try to implement this as per your example, or the Flapjack processor respectively.

butchwarns:
Following your suggestion, I also implemented a new parameter in the drop-down, which enables self-oscillating behavior. Resonance range is now restricted by default. Will create new PR once I sorted out the smoothing thing!

Thanks again for your feedback!

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
butchwarns: Hi Jatin,

I was able to implement what you suggested by using smoothedBufferValues in the processor. The fork is now updated for you to have a look, if you feel like it! Everything seems to work, just the safety limit on resonance (popup parameter) is now no longer implemented. I am still contemplating if there is a way of doing this without excessive per-sample if statements in the new processor. Do you by any chance have an idea?

-Butch

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
```diff
@@ -0,0 +1,135 @@
...
+    lp_cutoff_smooth.setParameterHandle (p.lp_cutoff_norm);
+    lp_resonance_smooth.setParameterHandle (p.lp_resonance_norm);
+
+    drive_smooth.prepare (sampleRate, samplesPerBlock);
+    drive_norm_smooth.prepare (sampleRate, samplesPerBlock);
+    hp_cutoff_smooth.prepare (sampleRate, samplesPerBlock);
+    hp_resonance_smooth.prepare (sampleRate, samplesPerBlock);
+    lp_cutoff_smooth.prepare (sampleRate, samplesPerBlock);
+    lp_resonance_smooth.prepare (sampleRate, samplesPerBlock);
+
```
jatinchowdhury18, line 39: For the resonance parameters, I wonder if we might also want to set the mapping functions to handle the "skew" operation in the smoother, instead of in the `filter.set_resonance()` call? This would mean that when the parameter is being smoothed, we would only need to call the skew function once per-block instead of once per-sample.

Basically we would do:
```cpp
hp_resonance_smooth.mappingFunction = [] (float val)
{
    return ladder_filter_utility::map_linear_normalized (val, 0.0, 4.0);
};
```

And then edit the `filter.set_resonance()` functions so they don't apply the skew a second time.

jatinchowdhury18: Actually, this would allow us to implement the Filter Mode parameter as well:
```cpp
hp_resonance_smooth.mappingFunction = [this] (float val)
{
    if (p.filter_mode_norm == 0.0f)
        val *= 0.5f; // less resonance in "normalized" mode
    return ladder_filter_utility::map_linear_normalized (val, 0.0, 4.0);
};
```
Or something like that... you probably have a better idea of how you would want that to work.

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
```diff
@@ -0,0 +1,135 @@
...
+    const int num_samples = buffer.getNumSamples();
+
+    for (int channel = 0; channel < buffer.getNumChannels(); ++channel)
+    {
+        float* data = buffer.getWritePointer (channel);
+
+        // Process high-pass
+
+        // If we need smoothing..
+        hp_cutoff_smooth.process (num_samples);
```
jatinchowdhury18, line 59: So the `SmoothedBufferValue::process()` calls should happen _outside_ the channels loop, since otherwise we'd be running the smoother on the same parameter twice per block.

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
jatinchowdhury18: And one more thing... I think we'll need a mapping function for `drive_smooth` also, since if I remember right, that parameter needed some kind of Decibel conversion to end up in the correct range?

src/processors/tone/ladder_filter/LadderParameters.h
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+private:
+    juce::AudioProcessorValueTreeState& vts;
+
+public:
+    //==============================================================================
+    // Atomic pointers to access raw (normalized) parameter values stored in the VTS
+
+    std::atomic<float>* drive_norm;
```
jatinchowdhury18, line 67: So in BYOD, we're mostly using `chowdsp::FloatParameter*` instead of `std::atomic<float>*` (as is the common thing with most JUCE plugins), since `chowdsp::FloatParameter` allows for non-destructive parameter modulation. It should be a pretty simple change though, the main difference would be:
```diff
- hp_cutoff_norm = vts.getRawParameterValue ("HP_CUTOFF");
+ ParameterHelpers::loadParameterPointer (hp_cutoff_norm, vts, "HP_CUTOFF");
```

butchwarns: Force of habit using `std::atomic<float>*`, I guess. Easy change indeed!

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
butchwarns: Absolutely right, done.

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
butchwarns: Hi Jatin, many thanks for your ideas on this! 

The resonance skew was already done in the smoother, but I moved the final denormalization call there too, which makes sense of course. 

I also moved the skew/denormalization functions out of the `LadderParameters` class and instead use lambdas now, as per your suggestion. The problem I was encountering before was the dependency of the resonance mapping on the `p.filter_mode_norm` value, which is obviously not working in a static method. Filter mode selection now works well.

The only other issue I was having was my unsuccessful attempt of making the `filter_mode_norm` anything other than a `std::atomic<float>*`, as this always resulted in an immediate crash when trying to access the value. If you feel this needs changing, please tell me. Any hint on how to use a `chowdsp::ChoiceParameter` here would be welcome in that case :)

Anyway, the processor seems to work fine now. Feel free to do your own checks, of course!
Thanks again.

-Butch

jatinchowdhury18 (approved):
Awesome, thanks for the updates, I think this PR is basically ready to merge! I left a couple of little comments, and the only other thing is that there's a few float conversion warnings that are coming up (https://github.com/Chowdhury-DSP/BYOD/actions/runs/4712945063/jobs/8379955617?pr=272), so if you could take care of those, that would be fantastic!

src/processors/tone/ladder_filter/HighPassLadder.h
```diff
@@ -0,0 +1,64 @@
...
+    double get_sample_rate() const;
+
+    //==============================================================================
+
+    // Set ladder filter cutoff frequency [Hz]
+    // Range: 20 Hz to 20480 Hz (10 octaves, approximate range of human hearing)
+    void set_cutoff (const double cutoff);
+
+    // Set ladder filter resonance [1]
+    // Range: 0.0 to 1.0
```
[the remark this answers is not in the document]
butchwarns: Good catch, should be 0.0 to 4.0 indeed. Will fix asap

src/processors/tone/ladder_filter/LadderParameters.cpp
```diff
@@ -0,0 +1,32 @@
...
+    using namespace ParameterHelpers;
+    auto params = createBaseParams();
+
+    createPercentParameter (params, "HP_CUTOFF", "HP CUT", 0.0f);
+    createPercentParameter (params, "HP_RESONANCE", "HP RES", 0.0f);
+    createPercentParameter (params, "LP_RESONANCE", "LP RES", 0.0f);
+    createPercentParameter (params, "LP_CUTOFF", "LP CUT", 1.0f);
+    createPercentParameter (params, "DRIVE", "DRIVE", 0.5f);
+
+    emplace_param<AudioParameterChoice> (params, "FILTER_MODE", "FILTER MODE", StringArray { "Normal", "Oscillating" }, 0);
```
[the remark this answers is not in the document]
butchwarns: Hmm, must have been me doing these changes late in the evening. Seems like exactly what I tried. Maybe I'll give it another try if I feel like it when back at the computer, thanks =)

butchwarns:
Fantastic. Thanks for checking again! I will, of course, take care of the float conversion warnings and other things you mentioned. Will let you know as soon as that's done. Cheers!

src/processors/tone/ladder_filter/LadderFilterProcessor.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include "LadderFilterProcessor.h"
+
+LadderFilterProcessor::LadderFilterProcessor (UndoManager* um) : BaseProcessor ("Ladder Filter", LadderParameters::createParameterLayout(), um), p (vts)
+{
+    // Info and style
+    uiOptions.info.description = "Resonant 4-pole ladder high-pass and low-pass,\n capable of self-oscillation. With nonlinear drive.";
+    uiOptions.info.authors = StringArray { "Butch Warns" };
+    uiOptions.info.infoLink = "https://butchwarns.de/";
+    uiOptions.backgroundColour = juce::Colour::fromRGB (0xBB, 0xE1, 0xC3);
+    uiOptions.powerColour = juce::Colour::fromRGB (0x99, 0x0B, 0x3C);
```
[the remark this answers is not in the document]
butchwarns: I updated the color scheme, looks better now afaic.

butchwarns:
Hi Jatin, 

I was able to implement the changes you recommended. No more float warnings, improved color scheme and fixed comments. 

Is there anything else you think needs to be reworked?

Best
-Butch

jatinchowdhury18:
Nope, I think this is good to go! Merging now...

Thanks for your contribution :)

butchwarns:
Wonderful. Glad to see this as part of the BYOD project! And thank you for all the help and feedback! =)

PR #310, opened 2023-06-23
Processor Port Handling 

RachelMaryamLocke:
New templated constructor for BaseProcessor allows const PortType vectors inputPortTypes and outputPortTypes to be initialised at the time of instantiation based on the InputPort and OutputPort emuns of their derived processors and specified inputPortMapper and outputPortMapper. A second deligating constructor creates derived processors without custom mapping requirements and initialises the BaseProcessors const PortType vectors to basic audio I/O.

Screenshot 2023-06-23 at 14 36 01
Screenshot 2023-06-23 at 14 38 37

### DIFF
--- a/src/processors/BaseProcessor.cpp
+++ b/src/processors/BaseProcessor.cpp
@@ -4,22 +4,17 @@
 
 BaseProcessor::BaseProcessor (const String& name,
                               ParamLayout params,
-                              UndoManager* um,
-                              int nInputs,
-                              int nOutputs) : JuceProcWrapper (name),
-                                              vts (*this, um, Identifier ("Parameters"), std::move (params)),
-                                              numInputs (nInputs),
-                                              numOutputs (nOutputs)
+                              UndoManager* um) : BaseProcessor (
+    name,
+    std::move (params),
+    BasicInputPort {},
+    BasicOutputPort {},
+    um,
+    [] (auto)
+    { return PortType::audio; },
+    [] (auto)
+    { return PortType::audio; })
 {
-    onOffParam = vts.getRawParameterValue ("on_off");
-
-    outputBuffers.resize (jmax (1, numOutputs));
-    outputBuffers.fill (nullptr);
-    outputConnections.resize (numOutputs);
-
-    inputBuffers.resize (numInputs);
-    inputsConnected.resize (0);
-    portMagnitudes.resize (numInputs);
 }
 
 BaseProcessor::~BaseProcessor() = default;
--- a/src/processors/BaseProcessor.cpp
+++ b/src/processors/BaseProcessor.cpp
@@ -1,6 +1,6 @@
 #include "BaseProcessor.h"
 #include "gui/pedalboard/editors/ProcessorEditor.h"
-#include "processors/netlist_helpers/NetlistViewer.h"
+#include "netlist_helpers/NetlistViewer.h"
 
 BaseProcessor::BaseProcessor (const String& name,
                               ParamLayout params,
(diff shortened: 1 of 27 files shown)


src/processors/BaseProcessor.h
```diff
@@ -186,6 +250,16 @@ class BaseProcessor : private JuceProcWrapper
...
+    enum class BasicInputPort
+    {
+        AudioInput,
+    };
+
+    enum class BasicOutputPort
+    {
+        AudioOutput,
+    };
+
```
jatinchowdhury18, line 262: It might make sense to add an empty `NullPort` enum class so that processors that have either no inputs (InputProcessor) or no outputs (Tuner, Oscilloscope, etc) can use it.

src/processors/utility/FreqBandSplitter.cpp
```diff
@@ -1,7 +1,7 @@
 #include "FreqBandSplitter.h"
 #include "../ParameterHelpers.h"
 
-FreqBandSplitter::FreqBandSplitter (UndoManager* um) : BaseProcessor ("Frequency Splitter", createParameterLayout(), um, 1, numOuts)
+FreqBandSplitter::FreqBandSplitter (UndoManager* um) : BaseProcessor ("Frequency Splitter", createParameterLayout(), um)
```
jatinchowdhury18, line 4:
```suggestion
FreqBandSplitter::FreqBandSplitter (UndoManager* um) : BaseProcessor ("Frequency Splitter", createParameterLayout(), um, BasicInputPort, OutputPort)
```

`FreqBandSplitter` already has its own `OutputPort` enum.

src/processors/BaseProcessor.h
```diff
@@ -40,16 +48,72 @@ struct ConnectionInfo
...
+                                                                                         inputPortTypes (base_processor_detail::initialisePortTypes<InputPort> (inputPortMapper)),
+                                                                                         outputPortTypes (base_processor_detail::initialisePortTypes<OutputPort> (outputPortMapper))
+    {
+        std::cout << "Creating processor: " << name << std::endl;
+        std::cout << "With input ports: " << std::endl;
+        for (size_t i = 0; i < inputPortTypes.size(); ++i)
+            std::cout << "    " << magic_enum::enum_name ((InputPort) i) << " (" << magic_enum::enum_name (inputPortTypes[i]) << ")" << std::endl;
+        std::cout << "With output ports: " << std::endl;
+        for (size_t i = 0; i < outputPortTypes.size(); ++i)
+            std::cout << "    " << magic_enum::enum_name ((OutputPort) i) << " (" << magic_enum::enum_name (outputPortTypes[i]) << ")" << std::endl;
```
jatinchowdhury18, line 101: This is great for confirming that everything works, but we should remember to remove it before merging :)

src/processors/BaseProcessor.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "JuceProcWrapper.h"
+#include "netlist_helpers/CircuitQuantity.h"
```
jatinchowdhury18, line 4: It would be great if we didn't need to include this inside this header... I'm curious what happens if we remove it?

src/processors/BaseProcessor.cpp
```diff
@@ -4,12 +4,16 @@
...
+    std::move (params),
+    BasicInputPort {},
+    BasicOutputPort {},
+    um,
+    [] (auto)
+    { return PortType::audio; },
+    [] (auto)
+    { return PortType::audio; })
 {
     onOffParam = vts.getRawParameterValue ("on_off");
```
jatinchowdhury18, line 18: Since this constructor already delegates to the other constructor, we no longer need to do anything in the constructor body.

src/processors/BaseProcessor.h
```diff
@@ -221,6 +295,8 @@ class BaseProcessor : private JuceProcWrapper
 
     juce::Array<int> inputModulationPorts {};
     juce::Array<int> outputModulationPorts {};
+    base_processor_detail::PortTypesVector inputPortTypes;
+    base_processor_detail::PortTypesVector outputPortTypes;
```
jatinchowdhury18, line 299: I think we can make these `const`?

src/processors/BaseProcessor.h
```diff
@@ -221,6 +295,8 @@ class BaseProcessor : private JuceProcWrapper
 
     juce::Array<int> inputModulationPorts {};
     juce::Array<int> outputModulationPorts {};
```
jatinchowdhury18, line 297: Eventually we should remove these, but maybe it's better to do that in the next PR?

src/processors/BaseProcessor.h
RachelMaryamLocke: Hey Jatin, I included this header as I was getting errors saying that the type was incomplete (having only the forward declaration in BaseProcessor) removing the include introduces the error `error: invalid application of 'sizeof' to an incomplete type 'netlist::CircuitQuantityList'` let me know if there is a better way around this!

src/processors/BaseProcessor.h
RachelMaryamLocke: Yep, for sure, was just leaving them for now whilst I work out what else needs to change to accommodate the portType vectors in the editor etc

src/processors/BaseProcessor.cpp
RachelMaryamLocke: yes, good point!

src/processors/BaseProcessor.h
RachelMaryamLocke: thanks, will remove before merging

PR #311, opened 2023-06-28
Level Ports

RachelMaryamLocke:
Refactoring of modulation port handling and introduction of more general port handling to accommodate level I/O ports

Screenshot 2023-06-28 at 08 47 03

### DIFF
--- a/src/gui/pedalboard/editors/ProcessorEditor.cpp
+++ b/src/gui/pedalboard/editors/ProcessorEditor.cpp
@@ -56,7 +56,7 @@ ProcessorEditor::ProcessorEditor (BaseProcessor& baseProc,
 
     for (int i = 0; i < baseProc.getNumInputs(); ++i)
     {
-        const auto portType = baseProc.isInputModulationPort (i) ? PortType::modulation : PortType::audio;
+        const auto portType = baseProc.getInputPortType (i);
         auto newPort = inputPorts.add (std::make_unique<Port> (procUI.backgroundColour, portType));
         newPort->setInputOutput (true);
         addAndMakeVisible (newPort);
@@ -64,7 +64,7 @@ ProcessorEditor::ProcessorEditor (BaseProcessor& baseProc,
 
     for (int i = 0; i < baseProc.getNumOutputs(); ++i)
     {
-        const auto portType = baseProc.isOutputModulationPort (i) ? PortType::modulation : PortType::audio;
+        const auto portType = baseProc.getOutputPortType (i);
         auto newPort = outputPorts.add (std::make_unique<Port> (procUI.backgroundColour, portType));
         newPort->setInputOutput (false);
         addAndMakeVisible (newPort);
--- a/src/headless/tests/BadModulationTest.cpp
+++ b/src/headless/tests/BadModulationTest.cpp
@@ -16,7 +16,7 @@ public:
 
             bool hasModulationInputs = false;
             for (int i = 0; i < proc->getNumInputs(); ++i)
-                hasModulationInputs |= proc->isInputModulationPort (i);
+                hasModulationInputs |= proc->getOutputPortType(i) == PortType::modulation;
 
             if (! hasModulationInputs)
                 processorsWithNoModulation.add (proc->getName());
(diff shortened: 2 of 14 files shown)


src/processors/BaseProcessor.cpp
```diff
@@ -356,20 +366,14 @@ juce::Point<int> BaseProcessor::getPosition (Rectangle<int> parentBounds)
...
 {
-    return inputModulationPorts.contains (portIndex);
+    return inputPortTypes[portIndex] == PortType::modulation;
 }
 
 bool BaseProcessor::isOutputModulationPort (int portIndex)
 {
-    return outputModulationPorts.contains (portIndex);
+    return outputPortTypes[portIndex] == PortType::modulation;
 }
```
jatinchowdhury18, line 377: Are we still using these methods? Since we now have `getInputPortType()` and `getOutputPortType()`, maybe we can remove these methods.

src/gui/pedalboard/editors/ProcessorEditor.cpp
```diff
@@ -56,15 +56,15 @@ ProcessorEditor::ProcessorEditor (BaseProcessor& baseProc,
 
     for (int i = 0; i < baseProc.getNumInputs(); ++i)
     {
-        const auto portType = baseProc.isInputModulationPort (i) ? PortType::modulation : PortType::audio;
+        const auto portType = baseProc.getInputPortType (i);
```
jatinchowdhury18, line 59: Now that feels better :)

jatinchowdhury18 (approved):
Looking good! Just a couple little notes.

src/processors/BaseProcessor.h
```diff
@@ -176,6 +176,8 @@ class BaseProcessor : private JuceProcWrapper
 
     int getNumInputs() const noexcept { return numInputs; }
     int getNumOutputs() const noexcept { return numOutputs; }
+    PortType getInputPortType (int portIndex);
+    PortType getOutputPortType (int portIndex);
```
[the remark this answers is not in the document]
RachelMaryamLocke: I made the return type const at first and got a warning saying `Clang-Tidy: Return type 'const PortType' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness` Does it make sense for them to be const at function level instead?

`PortType getInputPortType (int portIndex) const;
  PortType getOutputPortType (int portIndex) const;`

src/processors/BaseProcessor.cpp
RachelMaryamLocke: `isInputModulationPort` is only used in BadModulationTest.cpp which doesn't sound like something that's being used 😄 

`isOutputModulationPort` is being used in `isOutputModulationPortConnected` if we still want to keep `isOutputModulationPortConnected` I can change the lines: 

`if (isOutputModulationPort (info.startPort))`
 `return true;`
to
`if (getOutputPortType(info.startPort) == PortType::modulation)` 
                `return true;`

Or do we want the connection checks to be generalised for multiple port types?

src/processors/BaseProcessor.h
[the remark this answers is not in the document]
jatinchowdhury18: Yeah, `const` at the function level is what I meant.

src/processors/BaseProcessor.cpp
jatinchowdhury18: Yeah, if we could switch to using the new methods in both places, that would be great!

src/headless/tests/BadModulationTest.cpp
```diff
@@ -16,7 +16,7 @@ class BadModulationTest : public UnitTest
 
             bool hasModulationInputs = false;
             for (int i = 0; i < proc->getNumInputs(); ++i)
-                hasModulationInputs |= proc->isInputModulationPort (i);
+                hasModulationInputs |= proc->getOutputPortType (i) == PortType::modulation;
```
jatinchowdhury18, line 19:
```suggestion
                hasModulationInputs |= proc->getInputPortType (i) == PortType::modulation;
```

PR #316, opened 2023-07-06
Compressor Level Tracking

RachelMaryamLocke:
Methods that handle audio I/O processing updated to also handle level I/O processing


src/processors/other/Compressor.h
```diff
@@ -23,6 +24,8 @@ class Compressor : public BaseProcessor
     chowdsp::FloatParameter* makeupDBParam = nullptr;
 
     AudioBuffer<float> levelBuffer;
```
jatinchowdhury18, line 26: I think we can remove `levelBuffer` now, and just use `levelOutBuffer` instead?

PR #322, opened 2023-07-18
Level Detective Processor - Audio Processing & Visualiser

RachelMaryamLocke:
Screenshot 2023-07-25 at 00 51 08

### DIFF
--- a/src/processors/other/LevelDetective.cpp
+++ b/src/processors/other/LevelDetective.cpp
@@ -1,0 +1,166 @@
+#include "LevelDetective.h"
+#include "../ParameterHelpers.h"
+#include "gui/utils/ModulatableSlider.h"
+
+using namespace chowdsp::compressor;
+
+namespace
+{
+const auto powerColour = Colours::gold.darker (0.1f);
+const auto backgroundColour = Colours::teal.darker (0.1f);
+const String attackTag = "attack";
+const String releaseTag = "release";
+} // namespace
+
+LevelDetective::LevelDetective (UndoManager* um) : BaseProcessor (
+    "Level Detective",
+    createParameterLayout(),
+    InputPort {},
+    OutputPort {},
+    um,
+    [] (InputPort)
+    {
+        return PortType::audio;
+    },
+    [] (OutputPort)
+    {
+        return PortType::level;
+    })
+{
+    using namespace ParameterHelpers;
+    loadParameterPointer (attackMsParam, vts, attackTag);
+    loadParameterPointer (releaseMsParam, vts, releaseTag);
+    uiOptions.backgroundColour = backgroundColour;
+    uiOptions.powerColour = powerColour;
+    uiOptions.info.description = "A simple envelope follower";
+    uiOptions.info.authors = StringArray { "Rachel Locke" };
+}
+
+ParamLayout LevelDetective::createParameterLayout()
+{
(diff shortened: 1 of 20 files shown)


src/processors/other/LevelDetective.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include "../../modules/chowdsp_utils/modules/gui/chowdsp_visualizers/CompressorPlots/chowdsp_LevelDetectorVisualizer.h"
```
jatinchowdhury18, line 3: I don't think we should need this include? `<chowdsp_visualizers/chowdsp_visualizers.h>` should be included in JuceHeader.h, and which should have this header included internally.

jatinchowdhury18:
Looking good overall! I was thinking it might make sense to override `processAudioBypassed()`. In its current form, I think the bypass behaviour would just pass through the audio without going through the level detector. I think more likely the behaviour we would want would be a "silent" output if the processor bypassed. Would you agree?

RachelMaryamLocke:
Yes, I agree, it wouldn't make sense for the level output to send out mono audio information when the processor is bypassed, so in the case it is bypassed I have cleared the levelOutBuffer.

RachelMaryamLocke:
Let me know if I'm ok to merge this :)

src/processors/other/LevelDetective.cpp
```diff
@@ -0,0 +1,167 @@
...
+            for (auto* s : { &attackSlider, &releaseSlider })
+            {
+                s->setSliderStyle (Slider::RotaryHorizontalVerticalDrag);
+                s->setTextBoxStyle (Slider::TextBoxBelow, false, 80, 20);
+                s->setColour (Slider::textBoxHighlightColourId, powerColour.withAlpha (0.55f));
+                s->setColour (Slider::thumbColourId, powerColour);
+                addAndMakeVisible (s);
+            }
+
+            visualiser.backgroundColour = backgroundColour.darker (0.4f);
```
jatinchowdhury18, line 128: So to be able to push this change through, we'll need to update the submodule for chowdsp_utils. So the steps for that are:
- in your local copy create new branch and make a PR with your changes
- once the PR is merged, pull the submodule to the latest commit, and then push the submodule update to this repo

src/processors/other/LevelDetective.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+    else
+    {
+        levelOutBuffer.clear();
+    }
+
+    outputBuffers.getReference (LevelOutput) = &levelOutBuffer;
+}
+
+void LevelDetective::processAudioBypassed (AudioBuffer<float>& buffer)
```
jatinchowdhury18, line 86: At the moment the visualizer doesn't update when the module is bypassed... do you think that's the correct behaviour? I'm kind of on the fence.

src/processors/other/LevelDetective.cpp
RachelMaryamLocke: Is this different to the branch and PR I created for utils with the changes to waveformView? [https://github.com/Chowdhury-DSP/chowdsp_utils/pull/446]

jatinchowdhury18: Nope that's the one! Just merged it.

PR #325, opened 2023-07-27
Level Port Section Added to BYOD Manual

RachelMaryamLocke:


jatinchowdhury18 (approved):
Yeah this looks good! I'm not sure I'd consider "Crying Child" and "Envelope Filter" to be "dynamics processors" but I'm not sure what word to use there instead.

RachelMaryamLocke:
I have updated the copy to reflect this better!

PR #356, opened 2024-04-27
Create plugins.json

RustoMCSpit:
studiorack registry


RustoMCSpit:
btw this is a static build, DO NOT COMMIT THIS

PR #361, opened 2024-05-31
my bloody valentine preset

RustoMCSpit:
https://en.wikipedia.org/wiki/My_Bloody_Valentine_(band)

### DIFF
--- a/res/presets/my bloody valentine.chowpreset
+++ b/res/presets/my bloody valentine.chowpreset
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<Preset name="my bloody valentine" plugin="BYOD" vendor="User" category=""
+        version="1.3.0" preset_file="/home/henry/Sync/ThirdPartyStuff/.clap plugins/BYOD/userpresets/User/my bloody valentine.chowpreset">
+  <proc_chain state_plugin_version="1.3.0">
+    <Graphic_EQ>
+      <Parameters x_pos="0.1529999971389771" y_pos="0.02641509473323822">
+        <PARAM id="gain_0" value="4.799999237060547"/>
+        <PARAM id="gain_1" value="5.0"/>
+        <PARAM id="gain_2" value="-1.100000381469727"/>
+        <PARAM id="gain_3" value="7.200000762939453"/>
+        <PARAM id="gain_4" value="-1.800000190734863"/>
+        <PARAM id="gain_5" value="2.399999618530273"/>
+        <PARAM id="on_off" value="1.0"/>
+      </Parameters>
+      <port_0 connection_0="2" connection_end_0="0"/>
+    </Graphic_EQ>
+    <Amp_IRs>
+      <Parameters x_pos="0.6830000281333923" y_pos="0.03018867969512939">
+        <PARAM id="gain" value="4.75"/>
+        <PARAM id="ir" value="3.0"/>
+        <PARAM id="mix" value="0.449999988079071"/>
+        <PARAM id="on_off" value="1.0"/>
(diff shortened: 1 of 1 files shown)


RustoMCSpit:
also if you dont mind, can i merge request IRs?

res/presets/my bloody valentine.chowpreset
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<Preset name="my bloody valentine" plugin="BYOD" vendor="User" category=""
+        version="1.3.0" preset_file="/home/henry/Sync/ThirdPartyStuff/.clap plugins/BYOD/userpresets/User/my bloody valentine.chowpreset">
```
jatinchowdhury18, line 4: A couple things here...
- Who do you want to be listed as the preset "vendor"? It's totally cool to use your on name (or screen name).
- To follow the naming conventions used in the other preset, we would probably want the category to be "Players".
- For presets that are shipped with the plugin, I usually delete the `preset_file` field, since that will be meaningless on other users' computers.

Here's an example preset file if you want to have a look: https://github.com/Chowdhury-DSP/BYOD/blob/main/res/presets/Superdrag.chowpreset#L3

jatinchowdhury18 (commented):
Thanks for sharing... I think I prefer the first of the two presets, but I'll leave that decision up to you. I tried playing the chords/lead for "When You Sleep" through both presets... both sounded like the mbv "vibe" to me generally, although not exactly like the recordings, but that probably has more to do with my playing rather than the presets.

In order for the preset to be shipped with the plugin, there's a couple of changes that are required. First are some edits to the preset file itself, which are described in a comment. The other change is that we'll need to add the new preset to the list of presets in the code [here](https://github.com/Chowdhury-DSP/BYOD/blob/main/src/state/presets/PresetManager.cpp#L247) (note that the list is in alphabetical order).

Please feel free to contribute more presets if you would like to!

res/presets/my bloody valentine.chowpreset
RustoMCSpit: > Who do you want to be listed as the preset "vendor"? It's totally cool to use your on name (or screen name).

oh just list it as yourself for now, for the sake of convenience. put it under Players

> For presets that are shipped with the plugin, I usually delete the preset_file field, since that will be meaningless on other users' computers.

yeah do that lol

RustoMCSpit:
is there no way to nest categories? id like to have more than one under CHOW/Player/mybloodyvalentine

RustoMCSpit:
https://github.com/Chowdhury-DSP/BYOD/actions/runs/9341970549 what does this mean

jatinchowdhury18:
> is there no way to nest categories? id like to have more than one under CHOW/Player/mybloodyvalentine

At the moment, there's no way to nest categories... I suppose we could add that in the future.

That said, I'm not sure it makes sense to have two separate mbv presets, especially since the two are so similar.

> https://github.com/Chowdhury-DSP/BYOD/actions/runs/9341970549 what does this mean

I wouldn't worry about CI runs happening on your fork, I'll have the CI pipeline run on this PR in a minute.

RustoMCSpit:
> That said, I'm not sure it makes sense to have two separate mbv presets, especially since the two are so similar.

i could make more that sound different, it's the idea rather than the current presets

> > https://github.com/Chowdhury-DSP/BYOD/actions/runs/9341970549 what does this mean
> 
> I wouldn't worry about CI runs happening on your fork, I'll have the CI pipeline run on this PR in a minute.

whats a CI run lol

RustoMCSpit:
will this be merged?

PR #386, opened 2026-07-01
Feature/NAMsupport base

chrysophylax:
# Enhancement
This PR adds support for neural-amp modeler (NAM) models using the https://github.com/mikeoliphant/NeuralAudio library.

# Licence 
NeuralAudio is MIT

### DIFF
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -18,6 +18,71 @@ add_subdirectory(sst-cpputils)
 add_subdirectory(ea_variant)
 add_subdirectory(math_approx)
 
+# add NeuralAudio (for NAM processor). We can't use its top-level CMakeLists
+# directly because it re-adds RTNeural and math_approx via add_subdirectory,
+# which would conflict with the targets already defined above. Instead we
+# build NeuralAudio's sources ourselves and reuse the deps we already have.
+# The NeuralAudio processor is not built on iOS.
+if(NOT IOS)
+    set(NEURAL_AUDIO_DIR ${CMAKE_CURRENT_SOURCE_DIR}/NeuralAudio)
+    set(NAM_CORE_DIR ${NEURAL_AUDIO_DIR}/deps/NeuralAmpModelerCore)
+
+    set(NEURAL_AUDIO_SOURCES
+        ${NEURAL_AUDIO_DIR}/NeuralAudio/NeuralModel.cpp
+        ${NEURAL_AUDIO_DIR}/NeuralAudio/RTNeuralLoader.cpp
+    )
+
+    set(NAM_CORE_SOURCES
+        ${NAM_CORE_DIR}/NAM/activations.cpp
+        ${NAM_CORE_DIR}/NAM/conv1d.cpp
+        ${NAM_CORE_DIR}/NAM/get_dsp.cpp
+        ${NAM_CORE_DIR}/NAM/ring_buffer.cpp
+        ${NAM_CORE_DIR}/NAM/lstm.cpp
+        ${NAM_CORE_DIR}/NAM/dsp.cpp
+        ${NAM_CORE_DIR}/NAM/container.cpp
+        ${NAM_CORE_DIR}/NAM/wavenet/slimmable.cpp
+        ${NAM_CORE_DIR}/NAM/wavenet/model.cpp
+        ${NAM_CORE_DIR}/NAM/wavenet/a2_fast.cpp
+    )
+
+    add_library(NeuralAudio STATIC ${NEURAL_AUDIO_SOURCES} ${NAM_CORE_SOURCES})
+    target_compile_features(NeuralAudio PUBLIC cxx_std_20)
+    # Note: these definitions are PRIVATE so they don't leak into consumers
+    # (BYOD's own code uses a different RTNeural backend).
+    target_compile_definitions(NeuralAudio PRIVATE
+        NAM_SAMPLE_FLOAT
+        DSP_SAMPLE_FLOAT
+        BUILD_NAMCORE
+        NAM_ENABLE_A2_FAST
+        BUILD_INTERNAL_STATIC_WAVENET
+        WAVENET_MATH=FastMath
+        LSTM_MATH=FastMath
+        WAVENET_MAX_NUM_FRAMES=64
+        LAYER_ARRAY_BUFFER_PADDING=24
+        RTNEURAL_USE_EIGEN=1
+        RTNEURAL_DEFAULT_ALIGNMENT=16
+    )
+    target_include_directories(NeuralAudio
+        PUBLIC
+            ${NEURAL_AUDIO_DIR}
+        PRIVATE
+            ${NAM_CORE_DIR}
+            ${NEURAL_AUDIO_DIR}/deps/RTNeural
+            ${NEURAL_AUDIO_DIR}/deps/RTNeural/modules/Eigen
+            ${NEURAL_AUDIO_DIR}/deps/RTNeural/modules/json
+            ${NEURAL_AUDIO_DIR}/deps/math_approx/include
+    )
+    set_target_properties(NeuralAudio PROPERTIES
+        POSITION_INDEPENDENT_CODE TRUE
+        VISIBILITY_INLINES_HIDDEN TRUE
+        C_VISIBILITY_PRESET hidden
+        CXX_VISIBILITY_PRESET hidden
+    )
+    if (MSVC)
+        target_compile_options(NeuralAudio PRIVATE /bigobj)
+    endif()
+endif()
+
 include(cmake/WarningFlags.cmake)
 include(cmake/ConfigFlags.cmake)
 include(cmake/RuntimeSIMDLib.cmake)
@@ -60,6 +125,20 @@ if(BYOD_BUILD_CLAP)
             chowdsp::chowdsp_clap_extensions
             clap_juce_extensions
     )
+endif()
+
+if(NOT IOS)
+    # WHOLE_ARCHIVE is required so that NAMCore's static ConfigParserHelper
+    # initializers (in wavenet/model.cpp, lstm.cpp, convnet.cpp, linear.cpp,
+    # container.cpp) survive the final link. Without it, the linker discards
+    # those object files (they only contain a static registration and are not
+    # referenced by symbol), and runtime lookup fails with:
+    #   "No config parser registered for architecture: WaveNet"
+    # The generator expression propagates the flag to the final executable /
+    # shared library link, as required for it to actually take effect.
+    target_link_libraries(juce_plugin_modules PRIVATE
+        "$<LINK_LIBRARY:WHOLE_ARCHIVE,NeuralAudio>")
+    target_compile_definitions(juce_plugin_modules PUBLIC BYOD_HAS_NEURAL_AUDIO=1)
 endif()
 
 target_compile_definitions(juce_plugin_modules
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -61,6 +61,7 @@ if(NOT IOS)
         LAYER_ARRAY_BUFFER_PADDING=24
         RTNEURAL_USE_EIGEN=1
         RTNEURAL_DEFAULT_ALIGNMENT=16
+        _USE_MATH_DEFINES
     )
     target_include_directories(NeuralAudio
         PUBLIC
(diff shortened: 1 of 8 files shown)


chrysophylax:
Tested on my Linux64-intel, Standalone v.1.3.1 on Fedora Atomic with Wayland + Cosmic DE using Pipewire audio and a DNAFx GiT Mobile II unit + external Ibanez guitar.

jatinchowdhury18:
Thanks for the PR. There's a lot of code here, so it's going to take me a minute to review it properly. To help me get started would you mind sharing the general approach you're taking to loading/saving the models in the plugin? Screenshots of any new UI elements are always useful as well.

Also, I wanted to clarify why the new module is `#ifdef`'d out for iOS... Is that your preference, or is that because of a license from a third-party dependency?

chrysophylax:
Hi!

For the model loading, I tried to emulate the behaviour available in other modules such as Amp IR and GuitarML. I save only the path to the model so when restored we reload it or default to blank if the file is missing.

From what I've seen with NAM, most users (incl. me) have a file tree where they have large banks from say, various Tone3000 collections that they load in. They can be pretty large too. Some of the capture-models seems to be for sale/non-embeddable so I stuck with just path as a pointer, instead of trying to serialize the models (like in GuitarML).

I opened this PR as a discussion piece, I'm a very recent user of this program :-) and if there's a different "style" preferred, I'll adjust it to align with that. 

The UI design is minimal and up for improvement!
 
image


Finally, for the iOS piece: I have no Apple device I can borrow or test on, so I figured I'd just exclude iOS completely for my PR to avoid the overhead of writing for a platform I cannot personally test at all. That and the NeuralAudio library not having any instructions for building for mobile (only desktop linux/windows) .

jatinchowdhury18:
> For the model loading, I tried to emulate the behaviour available in other modules such as Amp IR and GuitarML. I save only the path to the model so when restored we reload it or default to blank if the file is missing.

Sounds good! I've gone back and forth between saving file binary data vs. file paths, but I agree that it makes sense to save paths here given the size of NAM model files.

> Finally, for the iOS piece: I have no Apple device I can borrow or test on, so I figured I'd just exclude iOS completely for my PR to avoid the overhead of writing for a platform I cannot personally test at all. That and the NeuralAudio library not having any instructions for building for mobile (only desktop linux/windows) .

That makes sense as well... I think it's okay to merge this PR without iOS support and then I (or someone else) can do some testing on iOS and make any necessary changes.

modules/CMakeLists.txt
```diff
@@ -18,6 +18,71 @@ add_subdirectory(sst-cpputils)
...
+        DSP_SAMPLE_FLOAT
+        BUILD_NAMCORE
+        NAM_ENABLE_A2_FAST
+        BUILD_INTERNAL_STATIC_WAVENET
+        WAVENET_MATH=FastMath
+        LSTM_MATH=FastMath
+        WAVENET_MAX_NUM_FRAMES=64
+        LAYER_ARRAY_BUFFER_PADDING=24
+        RTNEURAL_USE_EIGEN=1
+        RTNEURAL_DEFAULT_ALIGNMENT=16
```
jatinchowdhury18, line 63: I think we'll need to add `_USE_MATH_DEFINES` here as well. I think that the cause of [this CI failure](https://github.com/Chowdhury-DSP/BYOD/actions/runs/28688988362/job/85181905803?pr=386).

modules/CMakeLists.txt
```diff
@@ -18,6 +18,71 @@ add_subdirectory(sst-cpputils)
 add_subdirectory(ea_variant)
 add_subdirectory(math_approx)
 
+# add NeuralAudio (for NAM processor). We can't use its top-level CMakeLists
+# directly because it re-adds RTNeural and math_approx via add_subdirectory,
+# which would conflict with the targets already defined above. Instead we
+# build NeuralAudio's sources ourselves and reuse the deps we already have.
+# The NeuralAudio processor is not built on iOS.
```
jatinchowdhury18, line 25: Yeah the duplicate dependencies is a bit unfortunate... I wish there was a way we could just tell the `NeuralAudio` cmake config to use the versions of those libraries that are already being built.

src/CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ target_sources(BYOD PRIVATE
     processors/drive/king_of_tone/KingOfToneDrive.cpp
     processors/drive/mouse_drive/MouseDrive.cpp
     processors/drive/muff_clipper/MuffClipper.cpp
+    $<$<NOT:$<BOOL:${IOS}>>:${CMAKE_CURRENT_SOURCE_DIR}/processors/drive/nam/NAMProcessor.cpp>
```
jatinchowdhury18: Isn't this already covered by the `#if ! JUCE_IOS` in `NAMProcessor.cpp`?

src/processors/ProcessorStore.cpp
```diff
@@ -20,6 +20,9 @@
 #include "drive/junior_b/JuniorB.h"
 #include "drive/king_of_tone/KingOfToneDrive.h"
 #include "drive/mouse_drive/MouseDrive.h"
+#if ! JUCE_IOS
+#include "drive/nam/NAMProcessor.h"
```
jatinchowdhury18, line 23: I think this is already covered by some `#if`s as well.

src/processors/drive/nam/NAMProcessor.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+        if (qualityChanged && qualityAppliedThisBlock)
+            lastQualityApplied = newQuality;
+    }
+
+    calOutGain.setGainDecibels (calOutDB);
+    calOutGain.process (buffer);
+
+    outGain.setGainDecibels (outputGainParam->getCurrentValue());
+    outGain.process (buffer);
```
jatinchowdhury18: I think we should be able to combine these two gain objects into a single one? Something like:
```cpp
outGain.setGainDecibels (outputGainParam->getCurrentValue() + calOutDB);
outGain.process (buffer);
```

src/processors/drive/nam/NAMProcessor.cpp
```diff
@@ -0,0 +1,533 @@
...
+    bool qualityAppliedThisBlock = true;
+
+    // NeuralAudio models must never be given more samples than the max
+    // buffer size they were built with: exceeding it is unchecked in release
+    // builds, and per the NeuralAudio README, SetMaxAudioBufferSize() is not
+    // realtime-safe so we can't resize here. This can happen if a
+    // prepare()-time model reload failed while the block size grew (e.g. the
+    // model file was deleted and the oversampling factor was raised); in
+    // that case, bypass the models until a reload succeeds.
+    if (numSamples <= modelsMaxBlockSize)
```
jatinchowdhury18, line 305: How often does this happen in practice? Is this something we could show the user a warning about in the UI?

src/processors/drive/nam/NAMProcessor.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+    calInGain.setGainDecibels (calInDB);
+    calInGain.process (buffer);
+
+    // Only push the quality parameter into the model when it has actually
+    // changed AND the model reports the change is realtime-safe. Otherwise
+    // the new value takes effect at the next model reload.
+    const float newQuality = qualityParam->getCurrentValue();
+    const bool qualityChanged = newQuality != lastQualityApplied;
+    bool qualityAppliedThisBlock = true;
```
jatinchowdhury18, line 296: This pattern seems a little bit complicated... I wonder if there's a way to simplify it.

src/processors/drive/nam/NAMProcessor.cpp
```diff
@@ -0,0 +1,533 @@
...
+                              ? (inputCalDBuParam->getCurrentValue() - modelInputLevelDBu)
+                              : 0.0f;
+    const float calOutDB = applyCalibration ? calOutputAdjustDB : 0.0f;
+
+    // Signal chain: user-in -> cal-in -> model -> cal-out -> user-out -> DC block
+    inGain.setGainDecibels (inputGainParam->getCurrentValue());
+    inGain.process (buffer);
+
+    calInGain.setGainDecibels (calInDB);
+    calInGain.process (buffer);
```
jatinchowdhury18: Similarly here I think we should be able to combine these two gain processors.

src/processors/drive/nam/NAMProcessor.cpp
```diff
@@ -0,0 +1,533 @@
...
+            Logger::writeToLog ("NAM: prepare() model reload failed for " + cachedModelPath);
+        }
+    }
+}
+
+void NAMProcessor::processAudio (AudioBuffer<float>& buffer)
+{
+    const SpinLock::ScopedTryLockType lock { modelChangingMutex };
+    if (! lock.isLocked())
+        return; // model is being swapped; pass audio through unchanged this block
```
jatinchowdhury18, line 274: Do you have a rough sense of how long the lock is held for when a new model is being loaded?

src/processors/drive/nam/NAMProcessor.cpp
```diff
@@ -0,0 +1,533 @@
...
+    static const auto stringTodBuVal = [] (const juce::String& s)
+    { return s.getFloatValue(); };
+    emplace_param<chowdsp::FloatParameter> (
+        params,
+        NAMTags::inputCalTag,
+        "Input Cal",
+        juce::NormalisableRange<float> { 0.0f, 30.0f },
+        12.0f,
+        std::function<juce::String (float)> { dBuValToString },
+        std::function<float (const juce::String&)> { stringTodBuVal });
```
jatinchowdhury18, line 73: Is there a reason not to use `createGainDBParameter()` here?

src/CMakeLists.txt
chrysophylax: Yep, my bad.  :)

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: Just wanted the dBu label to match the unit 'dBu' instead of 'dB' (that's what the nam model calls this bit)

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: Yep

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: Very rare, I did some hunting with ASan to find this but haven't encountered it myself. I didn't want a heap corruption however, so I added in this guard.

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: Yep!

modules/CMakeLists.txt
chrysophylax: Adding and let's see if it works :)

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: Measured! Locks are held for roughly single digits to tens of microseconds on my laptop (Zen4 arch 8-core).

chrysophylax: Since we only really hold the lock while swapping, the loading is lock-free

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: If you prefer to simplify over keeping dBu, I'll make the change.

jatinchowdhury18: Keeping dBu is fine with me!

chrysophylax:
I added some missing includes that were causing the windows build to error out :-)

src/processors/drive/nam/NAMProcessor.cpp
chrysophylax: Agreed but I didn't think of anything yet :sweat_smile:

modules/CMakeLists.txt
```diff
@@ -18,6 +18,72 @@ add_subdirectory(sst-cpputils)
 add_subdirectory(ea_variant)
 add_subdirectory(math_approx)
 
+# add NeuralAudio (for NAM processor). We can't use its top-level CMakeLists
```
jatinchowdhury18, line 21: I think it might be more convenient to extract most of this out to a separate file (maybe `cmake/NeuralAudio.cmake`) and then we can just include that file here? Just to keep this file a little bit more focused on the overall setup of the plugin rather than the details of the third-party libraries.

jatinchowdhury18:
Looking good!

Don't worry about the auto-formatter CI for now... if the other CI checks pass then I think we can merge this.

chrysophylax:
I'm back from vacation :-) anything I should do?

